PR #9, opened 2026-05-04
Implement passes-storage: AGP, SQLCipher, Keystore, manifest XMLs

bittelc:
## Summary

Implementation follow-on to `wpass-9vv.3` (design) closing `wpass-fpj`. Wires AGP onto `passes-storage` and lands the trust-claim-bearing storage layer behind the contract from ADR 0002.

- AGP 9.0.1 + SQLCipher 4.5.4, minSdk 28, matching `passes-ui`.
- `AndroidKeystorePassKeyProvider`: AES-256 master in AndroidKeyStore (StrongBox preferred, TEE fallback, Software last), AES-GCM-wrapped 32-byte DB key. `KeyBacking` reported through `StorageTelemetryGuard.onKeyProviderInitialized` (ADR D2).
- Wrapped key envelope persists in a private `SharedPreferences` file. Note: ADR D2 specified `schema_meta` for envelope storage, but `schema_meta` is itself inside the SQLCipher-encrypted DB the envelope unlocks; storing the envelope outside the DB resolves the chicken-and-egg without weakening the trust claim (the envelope is itself ciphertext under a non-exportable Keystore master).
- Library-merged `res/xml/walt_passes_backup_rules.xml` + `walt_passes_data_extraction_rules.xml`, wired via the library `AndroidManifest.xml`'s `android:fullBackupContent` and `android:dataExtractionRules` (ADR D5).
- `BackupRulesAssertion.assertBackupRulesApplied(context)`: consumer-side check the merge actually applied. Uses reflection on `ApplicationInfo.fullBackupContent` / `dataExtractionRulesRes` since both are hidden public-API fields in recent SDKs.
- `SqlCipherPassRepository`: StateFlow seeded from disk; upsert/delete serialized through a write mutex; delete updates the StateFlow inside the same critical section, then emits `onPassDeleted` (ADR D6 — no undo, no soft-delete, no VACUUM).
- Internal `PassStore` boundary lets the public-surface contract tests run on Robolectric without SQLCipher native libs.

## Tests

- `PublicApiSurfaceTest`: still passes; no public surface changes.
- `SchemaDdlTest` (JVM, xerial sqlite-jdbc): verifies DDL creates the four tables + three indexes, the unique identity index rejects duplicate imports, and `ON DELETE CASCADE` removes child rows in `pass_images` / `pass_locales`.
- `PassRepositoryContractTest` (Robolectric): StateFlow seeds from disk; upsert / delete reflect in StateFlow; telemetry fires after the StateFlow update; unknown-id delete returns `IntegrityViolation` and emits no `onPassDeleted`; the public surface has no `undelete` / `restore` / `softDelete` / `archive` method.

`./gradlew check` green across `passes-core`, `passes-storage`, `passes-ui`.

## Test plan

- [ ] Walk the merged manifest in a consumer (`assembleRelease`) and confirm `android:fullBackupContent` / `android:dataExtractionRules` reference the walt resources.
- [ ] Confirm `walt_passes_backup_rules.xml` and `walt_passes_data_extraction_rules.xml` ship in the AAR.
- [ ] Smoke-test `SqlCipherPassRepository.create(context, AndroidKeystorePassKeyProvider.create(context).value, telemetryGuard)` against a real device under wpass-x5l.

## Follow-ups

- `wpass-x5l` — device-required `connectedAndroidTest` suite: bmgr Auto Backup verification, StrongBox/TEE matrix on emulator + physical device, work-profile install, KeyUnavailable surface across app-data clear.

Closes wpass-fpj.

### DIFF
--- a/docs/adr/0002-passes-storage.md
+++ b/docs/adr/0002-passes-storage.md
@@ -29,7 +29,11 @@ Rationale: a per-pass key scheme would add complexity (key directory, per-row un
 
 ### D2. The database key is wrapped, not derived
 
-The DB key is generated once at first open as 32 random bytes from `SecureRandom`, then wrapped (AES-256-GCM) with a master key resident in Android Keystore. The wrapped blob and its IV are stored in the `schema_meta` table. On subsequent opens, the wrapped blob is unwrapped via Keystore and the resulting raw bytes are passed to SQLCipher's `PRAGMA key`.
+The DB key is generated once at first open as 32 random bytes from `SecureRandom`, then wrapped (AES-256-GCM) with a master key resident in Android Keystore. The wrapped blob and its IV are persisted in a private SharedPreferences file (`is.walt.passes.storage.key_envelope`) outside the SQLCipher database itself. On subsequent opens, the wrapped blob is unwrapped via Keystore and the resulting raw bytes are passed to SQLCipher's `PRAGMA key`.
+
+The envelope MUST live outside the encrypted database it unlocks: storing it inside `schema_meta` would create a chicken-and-egg (you would need the DB key to read schema_meta, but the schema_meta entry IS the DB key). The earlier draft of this ADR placed the envelope in `schema_meta`; the implementation bead corrected this to a SharedPreferences file. The envelope is itself ciphertext under a Keystore-resident, non-exportable master, so the SharedPreferences file's lack of file-level encryption does not weaken the trust claim. The same `walt_passes_backup_rules.xml` / `walt_passes_data_extraction_rules.xml` rules exclude this preferences file from cloud backup and device-to-device transfer (the SharedPreferences domain is opted out alongside the database).
+
+The envelope is written synchronously (`commit()`, not async `apply()`) before the unwrapped DB key is handed to SQLCipher. A crash between an async write and the disk flush would leave SQLCipher pages encrypted with a key whose envelope was never persisted, irreversibly bricking the at-rest data on next launch.
 
 Why wrapped, not Keystore-derived directly: Android Keystore does not export raw symmetric key bytes for AES keys (the keys are non-exportable by design). SQLCipher requires a raw 32-byte key to derive its page key via PBKDF2. Wrapping is the bridge: the master key never leaves Keystore (so hardware-backing applies), and the unwrapped DB key is held in process memory only as long as the database is open.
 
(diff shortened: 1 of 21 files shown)


bittelc:
## Round 2 — review fixes (commit da440a2)

Addresses items 1–8 from the self-review:

**Critical**
1. `WrappedKeyStorage.write()` is now synchronous (`commit()`) and returns a Boolean. `AndroidKeystorePassKeyProvider.provideDatabaseKey()` refuses to hand out an unanchored DB key if the envelope did not reach disk — closes the brick scenario where a crash between async `apply()` and the flush would leave SQLCipher pages encrypted under a key whose envelope was lost.
2. `runIo` rethrows `CancellationException` before the catch-all so structured cancellation works.
3. `PassRepository.close()` added (idempotent). Calls after `close()` return `DatabaseLocked` (with telemetry) instead of using a stale handle.
4. Migrated to `net.zetetic:sqlcipher-android:4.6.1` from the deprecated `android-database-sqlcipher:4.5.4`. Imports follow to `net.zetetic.database.sqlcipher.*`, native lib loaded via `System.loadLibrary("sqlcipher")`.
5. ADR 0002 D2 amended to document SharedPreferences envelope location + synchronous-write requirement; D8 public-API list includes `BackupRulesAssertion`.

**High**
6. `PRAGMA cipher_compatibility = 4` now issued per the ADR sequence diagram.
7. Every `StorageResult.Failure` routes through a single emission helper. Previously `load/summaryOf/delete` `IntegrityViolation` paths were silent.
8. Unknown `signature_status_kind` values surface as `onMigrationRowDropped(UnknownEnumValue)` and the row is dropped, instead of throwing and surfacing as `Unknown(Other)`.

Bonus: added a schema-version-on-disk check before DDL so a downgraded build returns `StorageError.Unsupported` (ADR D7).

`./gradlew check` green across all three modules. New tests:
- `deleteOfUnknownIdReturnsIntegrityViolationAndEmitsFailure`
- `closeMakesSubsequentCallsReturnDatabaseLockedAndEmitsFailure`

Items deferred to follow-ups: `BackupRulesAssertion` reflection-fragility note, additional Robolectric SDK matrix, `WrappedKeyStorage` / `PassJsonCodec` round-trip tests, dead `Schema.MetaKeys.KEY_ALIAS`/`KEY_BACKING` constants.

bittelc:
## Round 3 — reviewer fixes (commit 7ca6908)

**(1) Backup rules now match the trust claim.** Both XMLs now also `<exclude domain="sharedpref" path="is.walt.passes.storage.key_envelope.xml" />`. The envelope was being backed up in violation of the WrappedKeyStorage.kt comment. Verified the merged `passes-storage-release.aar` carries the sharedpref exclude in both files.

**(2) Fresh `KeyGenerator` on StrongBox fallback.** `generateMasterKey` no longer reuses the same `KeyGenerator` instance after a failed `init()` (undefined per the Keystore contract). Collapsed the spec-builder duplication into a single `generateWithSpec(strongBoxBacked: Boolean)` helper.

**(3) `KeyUnavailable` (not `DatabaseCorrupt`) when envelope is lost but DB survives.** `AndroidKeystorePassKeyProvider` now takes a `databaseFileExists` probe; when the SharedPreferences envelope is missing AND `walt_passes.db` exists on disk, the provider returns `KeyUnavailable` instead of generating a fresh key (which would later surface as `DatabaseCorrupt` and confuse the wallet's user-facing remediation copy). New `AndroidKeystorePassKeyProviderTest` pins the contract.

`./gradlew check` green across all three modules.

PR #18, opened 2026-05-05
BackupRulesAssertion: validate by content, support tools:replace consumers (wpass-edb)

bittelc:
## Summary

- `BackupRulesAssertion` now validates by **content**, not resource identity. Opens whichever XML resource the merged manifest points at and confirms every entry in `REQUIRED_EXCLUDES` is present in every backup-relevant section. Fixes the assertion for consumers (walt-android included) that use `tools:replace` and mirror the library's required `<exclude>` entries into a consumer-owned XML.
- Outcome surface widened: `Applied | MissingExcludes | NoBackupRulesConfigured | RulesResourceUnreadable | PackageInfoUnavailable | FieldUnavailable`. `MissingExcludes` enumerates every section/entry gap. `allowBackup=false` short-circuits to `Applied`.
- ADR 0002 D5 documents the inherit-vs-mirror consumer postures and the content-not-identity validation contract.

Refs: `wpass-edb`. Unblocks `wpass-x5l` (device-instrumentation tests) which depends on the post-fix assertion shape.

## Test plan

- [x] Parser unit tests: full-backup-content, data-extraction-rules with both sections, mirror-with-extras, ignored `<include>` and unknown elements, unknown root error, stray `<exclude>` outside any section.
- [x] End-to-end Robolectric test against the library's own `walt_passes_backup_rules.xml` and `walt_passes_data_extraction_rules.xml` returns `Applied`.
- [x] `allowBackup=false` short-circuit returns `Applied` without parsing rules.
- [x] Exhaustive `when` over every `Outcome` arm (compile-lock against silent surface drift).
- [x] `./gradlew check` green across all three modules (test, detekt, ktlint, lint).
- [ ] On-device verification (StrongBox/TEE matrix, bmgr backup-blob inspection): deferred to `wpass-x5l`.

### DIFF
--- a/docs/adr/0002-passes-storage.md
+++ b/docs/adr/0002-passes-storage.md
@@ -110,6 +110,13 @@ The library does NOT set `android:allowBackup="false"` on the consumer, because 
 
 A consumer-side `assertBackupRulesApplied()` helper is exposed from `passes-storage` for use in instrumentation tests, so the trust claim is checkable in CI rather than relying on documentation review.
 
+Two consumer postures are valid and the assertion accepts both:
+
+1. **Inherit.** The consumer lets the library's manifest contributions reach the merged manifest unchanged. `ApplicationInfo.fullBackupContent` and `ApplicationInfo.dataExtractionRulesRes` then reference `walt_passes_backup_rules` and `walt_passes_data_extraction_rules` directly.
+2. **Mirror.** The consumer overrides the library contributions with `tools:replace="android:fullBackupContent,android:dataExtractionRules"` and points the manifest at a consumer-owned XML resource that mirrors the library's required `<exclude>` entries (and optionally adds its own). Walt-android takes this posture so it can manage backup posture for the whole app from a single resource.
+
+The assertion validates by content, not resource identity. It opens whichever XML resource the merged manifest points at and checks that every entry in `BackupRulesAssertion.REQUIRED_EXCLUDES` is present in every backup-relevant section of that resource (`<full-backup-content>` for the API 23 - 30 path; both `<cloud-backup>` and `<device-transfer>` for the API 31+ path). Consumers may add additional excludes; only the pass-related entries are required for the trust claim. Setting `android:allowBackup="false"` app-wide trivially satisfies the claim and short-circuits the assertion.
+
 ### D6. Deletion is items 1+3+4+5+6 from `decision-wlt-0tn-q3-4`
 
 `PassRepository.delete(id)`:
--- a/passes-storage/src/main/kotlin/is/walt/passes/storage/BackupRulesAssertion.kt
+++ b/passes-storage/src/main/kotlin/is/walt/passes/storage/BackupRulesAssertion.kt
@@ -3,73 +3,281 @@ package `is`.walt.passes.storage
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.content.pm.PackageManager
+import android.content.res.Resources
+import android.content.res.XmlResourceParser
 import android.os.Build
+import org.xmlpull.v1.XmlPullParser
+import org.xmlpull.v1.XmlPullParserException
+import java.io.IOException
 
 /**
- * Verifies that the consumer's merged `AndroidManifest.xml` actually carries the
- * library-shipped backup-exclusion rules from ADR 0002 D5. Intended for the consumer's
- * instrumentation test suite; the trust claim "pass data is excluded from cloud backup"
- * becomes checkable in CI rather than relying on manifest review.
+ * Verifies that the trust claim "pass data is excluded from cloud backup" holds in the
+ * consumer's merged manifest. Intended for the consumer's instrumentation test suite so
+ * that the claim is checkable in CI rather than relying on manifest review.
  *
- * The assertion only checks that the consumer's `<application>` references rule resources
- * named `walt_passes_backup_rules` (API 23-30) and `walt_passes_data_extraction_rules`
- * (API 31+). It does NOT parse the rules XML. Combined with the resources we ship, this
- * narrows the consumer-side risk to "the consumer overrode our rules with a non-walt
- * resource of the same name," which is a deliberate act, not an oversight.
+ * The library ships `walt_passes_backup_rules.xml` (API 23 - 30) and
+ * `walt_passes_data_extraction_rules.xml` (API 31+). Two consumer postures are valid:
  *
- * The two ApplicationInfo fields we check (`fullBackupContent` and `dataExtractionRulesRes`)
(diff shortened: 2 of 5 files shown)


bittelc:
Pushed 64a9bc2 addressing the review. Each point individually assessed; here is the disposition:

**Done in this commit (10 of 14):**

- **#1 shared `ApplicationInfo` mutation** — added internal `@VisibleForTesting` overload `assertBackupRulesApplied(ApplicationInfo, Resources)`; the `allowBackup=false` test now Parcel-clones `ApplicationInfo` and never touches Robolectric's shared instance.
- **#2 opaque "parse failed" diagnostics** — `parseResource` catches `BackupRulesParseException`, `XmlPullParserException`, `IOException`, and `Resources.NotFoundException` distinctly and propagates `e.message` into `RulesResourceUnreadable.reason`. A consumer pointing at `network_security_config.xml` now sees `"unexpected root element: network-security-config"`.
- **#3 `error()` for control flow** — replaced with a typed internal `BackupRulesParseException`. No more generic `IllegalStateException` coercion.
- **#5 `validateResource` mixed mutation + return** — now returns a private sealed `ValidationResult.Ok | Unreadable`; `assertBackupRulesApplied` reduces both validations explicitly.
- **#6 ParseState flag combinatorics** — replaced with root-dispatched strategies. `parseRulesXml` reads the first `START_TAG` once and dispatches to `parseDataExtractionRules` (DXR mode) or a single-section reader (legacy mode). No more `rootSeen` / `rootIsFullBackup` / `currentSection` state machine.
- **#7 scattered XML literals** — `xmlElement` attached to the `Section` enum. Unscoped element names (`data-extraction-rules`, `exclude`) and attribute names (`domain`, `path`) are file-private constants.
- **#8 `enter()` pre-creates dead-code set** — gone with the new parser shape.
- **#9 `@JvmStatic` on internal parser** — dropped.
- **#10 unnecessary `problems.toList()`** — dropped.
- **#11 extract parser to its own file** — three files now: `BackupRulesContract.kt` (the trust contract), `BackupRulesXmlParser.kt` (internal, root-dispatched), `BackupRulesAssertion.kt` (policy + Outcome).
- **#12 `REQUIRED_EXCLUDES` prominence** — moved to `BackupRulesContract` object alongside the section/exclude types. An auditor reading the trust claim now lands on a 25-line file dedicated to it.
- **smaller-doc** — KDoc says "Robolectric or on-device instrumentation" not "instrumentation".

**Filed as follow-up beads (3 of 14):**

- **#4 greylist → blocklist regression** — `wpass-2zv` (P3): investigate manifest-XML fallback if either `fullBackupContent` or `dataExtractionRulesRes` reflection breaks. Trigger: `Outcome.FieldUnavailable` observed on a tested API level.
- **#13 `Outcome.EnvironmentError` grouping** — `wpass-891` (P4): defensible refinement, defer until a third "environment" arm is needed or a consumer asks to discriminate the two categories.
- **smaller-deprecation `getApplicationInfo(String, Int)`** — `wpass-sbf` (P4): swap to `ApplicationInfoFlags` overload at the next minSdk bump.

**Declined (1 of 14):**

- **#14 `outcomeArmsAreReachableViaWhen` test is redundant** — kept. The pattern is the deliberate project convention from `PublicApiSurfaceTest.kt:9-12`: every sealed arm reached via exhaustive `when` so adding/removing an arm forces a compile-time conversation inside this library, before any consumer is touched. Added a doc comment to the test pointing at the convention.

`./gradlew check` green across all three modules.

bittelc:
Pushed 7cbab46 addressing the re-review. Disposition:

**Done in this commit (7 of 9):**

- **#1 pre-S branch untested** — added `@Config(sdk = [30])` end-to-end test (`endToEndOnPreSReturnsAppliedFromFullBackupContentAlone`). Exercises `readDxrResId`'s `else 0` branch on Android 11 where `ApplicationInfo.dataExtractionRulesRes` does not exist.
- **#2 endToEnd path-ambiguous** — now asserts `allowBackup=true` before `Applied` so a manifest flip can't silently make the test pass via short-circuit.
- **#3 duplicate-section overwrite** — switched to union (`sections[section]?.let { it + collected } ?: collected`). Test pinned (`duplicateSectionsAreUnioned`). A missing entry in a first `<cloud-backup>` block can't be masked by its presence in a duplicate block.
- **#4 depth-blind collection** — documented the deliberate tolerance in `BackupRulesXmlParser` KDoc alongside the duplicate-section policy. The Android schema disallows nesting, so a malformed document with nested elements is the consumer's bug to fix; the assertion does not try to second-guess what the OS itself would reject.
- **#5 top-level parser** — wrapped in `internal object BackupRulesXmlParser`. Exception renamed to `BackupRulesXmlParser.ParseException`. Symmetric with the other two files.
- **#7 awkward `as` casts in `reduce`** — switched to elvis-collapsed `as?` form. Two returns instead of three (detekt limit).
- **#8 public entry uncovered for `allowBackup=false`** — same `endToEnd` test now cross-checks the public entry point against the seam on the same context. Catches `Context.applicationInfo` vs `packageManager.getApplicationInfo(packageName)` divergence under any future Robolectric configuration. Full coverage of the public allowBackup=false path would re-introduce the shared-state hazard the seam was designed to avoid; consistency check is the cheapest substitute.
- **#9 surface-lock convention** — KDoc pointer added on `sectionEnumerates…` so future maintainers see the same `PublicApiSurfaceTest` convention already referenced from `outcomeArmsAreReachableViaWhen`.

**Deferred (1 of 9):**

- **#6 type indirection collapse** (ResourceIds / ParseAttempt / ValidationResult) — reviewer explicitly flagged "not blocking; flag for a future cleanup if the file grows." Skipped without filing a bead; the trigger is file growth, which we'll see in the diff if it happens.

`./gradlew check` green across all three modules.

bittelc:
Pushed 05bd791 addressing the third-pass review.

**Done (3 of 4):**

- **#1 pre-S endToEnd not hardened** — mirrored the API 34 shape: `FLAG_ALLOW_BACKUP != 0` precondition + public-vs-seam cross-check on the same context.
- **#2 `reduce` reports only first `Unreadable`** — added the one-line comment as suggested. The single-resourceId `Outcome` shape doesn't model simultaneous failures; consumer fixes the first and the second surfaces on the next run.
- **#3 `Map.merge` over the elvis chain** — swapped. Equivalent semantics, more direct.

**Skipped (1 of 4):**

- **#4 `ApplicationInfo.CREATOR` non-null on Robolectric** — reviewer flagged "not blocking." `CREATOR` is part of the stable Parcelable surface; if Robolectric ever regresses, the failure would be loud, not silent. No action.

`./gradlew check` green.

PR #24, opened 2026-05-05
passes-core: hardened ZIP extraction layer (wpass-8so)

bittelc:
## Summary

First slice of the `PassParser` implementation: an internal `extractSafely()` entry point that takes a `PassSource` and returns either a `Map<String, ByteArray>` of safe entries or a `MalformedReason`. Every guard the threat model lists for untrusted PKPASS input is centralized here before any bytes are surfaced to a downstream parser.

`ExtractResult` and `extractSafely` are package-private under `is.walt.passes.core.internal`, so the public API surface from ADR 0001 stays frozen and `PublicApiSurfaceTest` is unchanged.

## Guard surface

- **Magic-byte preflight.** `ZipInputStream` silently treats unrecognized leading bytes as 'no entries', which would let raw garbage and 0-byte input round-trip as `Success(empty)`. The first 4 bytes are sniffed and anything that isn't `PK\x03\x04` (local file header) or `PK\x05\x06` (end-of-central-directory; legitimately empty archive) is rejected.
- **Archive size** (compressed). Checked twice: up-front against the declared size from `PassSource.Bytes` length / `PassSource.Stream.sizeHintBytes`, then again at streaming time via a `BoundedInputStream` that throws on overflow. The streaming check is load-bearing because a hostile `Stream` source can lie about its declared size.
- **Entry count** capped via `ParserConfig.maxEntries`. Directory entries are skipped before counting so a bag of nested `.lproj/` directories cannot push a real archive past the cap.
- **Per-entry decompressed size** capped via `ParserConfig.maxEntryBytes`. This is the zip-bomb guard: a 10 KB compressed entry that decompresses to 10 GB hits the cap and aborts before the buffer materializes.
- **Path traversal (zip-slip).** Rejects `..` and `.` segments, leading `/`, backslashes, Windows drive-letter prefixes, and empty path segments. Structural; no file-system canonicalization, because we never touch the file system.
- **Extension allowlist.** Entry names must end in `.json`, `.png`, or `.strings`, OR be exactly `signature` at the archive root (the PKCS#7 detached-signature blob has no extension by PKPASS convention; the exemption is intentionally root-only so an attacker can't smuggle arbitrary content under a nested `signature` name).
- **Duplicate entry names** rejected. PKPASS does not permit duplicates and `ZipInputStream` would silently let the second one win, which would let an attacker shadow a legitimate `manifest.json` with a tampered second copy.
- **In-memory only.** No `FileOutputStream` is ever opened on an entry name. The entire archive is materialized into a `Map<String, ByteArray>` bounded by the per-entry cap.

## Out of scope (follow-up beads)

- `PassParser.create()` factory wiring (parser-glue bead).
- `pass.json` deserialization, manifest hash chain, PKCS#7 signature verification, `.strings` parsing.
- Coroutine timeouts around parsing.
- Dedicated `MalformedReason` arm for structural rejections (path traversal / disallowed extension / duplicate entry); currently surfaced as `MalformedReason.NotAZipArchive` because the public surface is frozen for this slice. Tracked in **wpass-n6g**.
- Explicit symlink rejection. JDK-only zip APIs don't expose the file-mode bits used to detect Info-ZIP symlinks (those live in the central directory, which `ZipInputStream` never reads). In-memory-only extraction + path-traversal + extension allowlist are sufficient for the trust claim today; **wpass-n6g** also captures the option to swap in a parser that exposes external attributes.

## Test plan

- [x] `./gradlew :passes-core:check` (test + ktlint, both green)
- [x] `./gradlew :passes-core:detekt` green
- [x] 25 new table-driven tests in `SafeArchiveExtractorTest` cover every guard above (happy path, all four resource limits, every path-traversal shape, disallowed extension, duplicate entry, signature-no-extension allowance, nested-signature rejection, magic-byte preflight, stream lifecycle).
- [x] Existing `PublicApiSurfaceTest` (13 cases) still passes unchanged - public surface is frozen.

Closes wpass-8so. Follow-up: wpass-n6g.

### DIFF
--- a/passes-core/src/main/kotlin/is/walt/passes/core/internal/ExtractResult.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/internal/ExtractResult.kt
@@ -1,0 +1,20 @@
+package `is`.walt.passes.core.internal
+
+import `is`.walt.passes.core.MalformedReason
+
+/**
+ * Outcome of running [extractSafely] over a [`is`.walt.passes.core.PassSource]. Internal
+ * only: the parser-glue bead lifts a [Failure] into
+ * [`is`.walt.passes.core.ParseResult.Malformed]. Surfacing this type to consumers would
+ * expand the public API beyond what ADR 0001 fixes.
+ */
+internal sealed interface ExtractResult {
+    /**
+     * Insertion-ordered map of entry name to fully-decoded entry bytes. Order mirrors the
+     * order the entries appeared in the archive's local-file-header stream so the
+     * downstream pass.json / manifest hash steps can iterate deterministically.
+     */
+    data class Success(val entries: Map<String, ByteArray>) : ExtractResult
+
+    data class Failure(val reason: MalformedReason) : ExtractResult
+}
--- a/passes-core/src/main/kotlin/is/walt/passes/core/internal/SafeArchiveExtractor.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/internal/SafeArchiveExtractor.kt
@@ -163,25 +163,35 @@ private fun validateAndRead(
     entries: MutableMap<String, ByteArray>,
     config: ParserConfig,
 ): ExtractResult.Failure? {
-    val rejection: MalformedReason? =
+    val rejection =
         pathTraversalReason(name)
-            ?: if (!hasAllowedName(name)) {
-                MalformedReason.NotAZipArchive
-            } else {
-                null
-                    ?: if (entries.containsKey(name)) {
-                        MalformedReason.NotAZipArchive
-                    } else {
-                        null
-                            ?: if (entries.size >= config.maxEntries) {
-                                MalformedReason.ResourceLimitExceeded(ResourceLimit.EntryCount)
-                            } else {
-                                null
-                            }
-                    }
-            }
+            ?: extensionReason(name)
+            ?: duplicateEntryReason(entries, name)
+            ?: entryCountReason(entries, config)
     return rejection?.let { ExtractResult.Failure(it) }
         ?: readEntryAndStore(zis, name, entries, config.maxEntryBytes)
+}
+
+private fun extensionReason(name: String): MalformedReason? {
+    return if (hasAllowedName(name)) null else MalformedReason.NotAZipArchive
+}
+
+private fun duplicateEntryReason(
+    entries: Map<String, ByteArray>,
+    name: String,
+): MalformedReason? {
+    return if (entries.containsKey(name)) MalformedReason.NotAZipArchive else null
+}
+
+private fun entryCountReason(
+    entries: Map<String, ByteArray>,
+    config: ParserConfig,
+): MalformedReason? {
+    return if (entries.size >= config.maxEntries) {
+        MalformedReason.ResourceLimitExceeded(ResourceLimit.EntryCount)
+    } else {
+        null
+    }
 }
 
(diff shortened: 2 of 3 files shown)


bittelc:
Thanks for the review. Pushed `5e34909` addressing points 1 and 3; point 2 already has a follow-up bead.

1. **`validateAndRead` style.** Refactored to the flat `?:` chain you sketched. The earlier nested form was an accident of getting the file under detekt's `TooManyFunctions` threshold by inlining helpers; with the file count now comfortable I re-extracted `extensionReason`, `duplicateEntryReason`, and `entryCountReason` so the call site reads top-to-bottom:

   ```kotlin
   val rejection = pathTraversalReason(name)
       ?: extensionReason(name)
       ?: duplicateEntryReason(entries, name)
       ?: entryCountReason(entries, config)
   ```

2. **`NotAZipArchive` overload.** Already filed as `wpass-n6g`. Will land before the parser-glue bead so consumer-facing code never sees the conflated arm.

3. **`readEntryBytes` keeps reading after deciding to fail.** Fixed - now returns the moment `totalRead > maxEntryBytes`. Still bounded by `BoundedInputStream` regardless, but no more wasted decompression on a near-cap payload.

Quality gates green: `:passes-core:check`, `:passes-core:detekt`. 25 extractor tests + 13 PublicApiSurface tests still pass; public surface unchanged.

bittelc:
Pushed `b49a629` for the second-pass items.

**In this PR:**

1. **Wrapper-order invariant doc.** Added a comment in `runZipPipeline` spelling out why `BoundedInputStream` must sit outside `BufferedInputStream` — bytes the magic-byte sniff prefetches still need to flow through the cumulative counter. Reordering would silently leak the first 4 (or however many the buffer pulled) past the `maxArchiveBytes` cap.
2. **EOCD comment in `looksLikeZip`** noting the second magic only legitimately matches an empty archive; anything else gets caught by `ZipInputStream` on the first read.
3. **Defensive directory path check.** Moved `pathTraversalReason` ahead of the `isDirectory` short-circuit in `processEntry` so a future change that starts acting on directory entries can't accidentally bypass the guard. Adjusted `pathTraversalReason` to strip one trailing `/` so legitimate `en.lproj/` directory names pass; intermediate empty segments (`foo//bar`) and a bare `/` still fail. New test `zipSlipDirectoryEntryIsRejected` pins this.
4. **Failure-path stream-open test.** `streamSourceFailurePathLeavesUnderlyingStreamOpen` now locks down the `NonClosingInputStream` contract on the abort path, not just success.

**Follow-up beads filed (deliberately out of scope for this PR):**

- **wpass-m8v** — route encrypted / unsupported-compression-method zips to `UnsupportedReason.EncryptedArchive` instead of folding into `NotAZipArchive`. Closes the dead-code observation on `ParseResult.UnsupportedReason.EncryptedArchive`.
- **wpass-oj8** — add `ParserConfig.maxTotalDecompressedBytes` so a future consumer loosening `maxEntries` doesn't silently raise the decompressed-footprint ceiling. Coordinates with **wpass-n6g** since `ResourceLimit` is on the public surface.

The minor items the reviewer marked "probably not worth doing" (hidden-filename `.json`, `entries.toMap()` defensive copy) are deliberately not touched. The "duplicate masks entry-count" ordering observation is a thing-to-know once dedicated `MalformedReason` arms land in **wpass-n6g**.

Test count: 27 extractor + 13 PublicApiSurface, all green. Quality gates green.

PR #25, opened 2026-05-05
passes-core: manifest.json hash chain verification (wpass-7l2)

bittelc:
## Summary

Second implementation slice of `PassParser`, building on the safe-extraction layer landed in #24 (wpass-8so). Adds an internal `verifyManifest(entries)` entry point that checks every file in the extracted archive against its SHA-1 hash declared in `manifest.json`, plus the structural rules PKPASS imposes around the manifest itself.

Tracks `wpass-7l2`. Blocks `wpass-dw2` (PKCS#7 signature verification) and `wpass-qnc` (parser-glue factory). `PassParser.create()` continues to throw until the glue bead.

## What changes

- `internal fun verifyManifest(entries: Map<String, ByteArray>): ManifestVerifyResult` — pure function, no I/O, returns either `Ok(manifestBytes)` (the bytes the signature bead will sign over) or `Failed(ManifestFailure)`.
- `ManifestFailure` is finer-grained than the frozen public `MalformedReason` surface so that `wpass-n6g` (dedicated `MalformedReason` arms) and `wpass-dw2` can route each shape independently.
- `ManifestHex.kt` carries the SHA-1 hex parser in its own file (kept out of `ManifestVerifier.kt` so the latter sits below the `TooManyFunctions` threshold).

## Failure-arm mapping

`ManifestFailure` arms map to `ParseResult` like this (mapping itself lands in `wpass-qnc`; the public surface remains frozen for this slice):

| `ManifestFailure`         | `ParseResult` arm                                         |
|---------------------------|-----------------------------------------------------------|
| `Missing`                 | `Malformed(MissingManifest)`                              |
| `InvalidJson`             | `Malformed(InvalidManifest)`                              |
| `InvalidShape`            | `Malformed(InvalidManifest)`                              |
| `InvalidHashFormat(name)` | `Malformed(InvalidManifest)`                              |
| `SelfReferentialEntry`    | `Malformed(InvalidManifest)`                              |
| `ExtraEntry(name)`        | `Malformed(InvalidManifest)`                              |
| `MissingEntry(name)`      | `Malformed(InvalidManifest)`                              |
| `HashMismatch(name)`      | `Tampered(FileHashMismatch)` *(security event, not malformedness)* |

Critically, `HashMismatch` is **never** coalesced with structural malformedness. The trust UI must distinguish tampering from "your file is broken".

## SHA-1 — why it's here, why it's not a problem

PKPASS uses SHA-1 in `manifest.json` per Apple's spec. This is **not** a security cipher choice: the actual cryptographic binding is the PKCS#7 detached signature over `manifest.json` bytes, which `wpass-dw2` validates. SHA-1 here is the structural hash chain; switching to SHA-256 would break every PKPASS writer in the wild (Apple Wallet, every issuer toolchain, every existing pass file). `Ok.manifestBytes` is returned verbatim from the entries map so the signature bead can hash and verify the exact bytes the envelope was constructed over.

## Order-of-checks invariant (load-bearing)

Within the per-entry loop: `InvalidHashFormat` -> `SelfReferentialEntry` -> `MissingEntry` -> `HashMismatch`. After the loop: `ExtraEntry`. The loop short-circuits on first failure, so `HashMismatch` (raised inside the loop) wins over `ExtraEntry` (raised after the loop) when both could fire. This matters once `wpass-n6g` lands dedicated `MalformedReason` arms — telemetry on hash-mismatch is operationally distinct from "structurally malformed", and an attack that simultaneously tampers a hash and adds a stray file should surface as the security event, not as the structural one.

## Test plan

- [x] `./gradlew :passes-core:check` (test + ktlintCheck)
- [x] `./gradlew :passes-core:detekt`
- [x] `PublicApiSurfaceTest` unchanged (no public-API mutation in this slice)
- [x] 22 in-memory `ManifestVerifierTest` cases — every `ManifestFailure` arm, hex case mixes, hex format rejections (wrong length, non-hex char), `signature` and `manifest.json` exemptions, empty manifest behavior, first-mismatch-wins ordering invariant


bittelc:
Addressed in `b32e2de`.

**Order swap (`SelfReferentialEntry` > `InvalidHashFormat`).** Done. Structural rule is the stronger statement; the manifest-line `"signature": "<bad-hex>"` case now surfaces `SelfReferentialEntry`. New regression test `selfReferentialEntryWinsOverInvalidHashFormat`.

**Iteration-order contract.** Documented on `verifyManifest`'s KDoc. The function consumes `entries` in iteration order; `extractSafely` returns an insertion-ordered map mirroring the local-file-header order, so calling code gets deterministic failure naming. A caller passing a `HashMap` (e.g. fuzz harness) gets non-deterministic *names*; correctness of accept-vs-reject is unaffected.

**Helper chain collapse.** 12 helpers -> 7. Depth 7 -> 4. Dropped the trivial 1-line wrappers (`checkManifest`, `parseJsonOrNull`, `readStringMap`, `collectStringEntries`, `firstEntryFailure`, `isManifestExempt`). The reviewer's two-function shape conflicts with the project's `detekt.yml` `ReturnCount=2` (each guard counts; `excludeGuardClauses=false`); the helpers that remain each have at most 2 returns and pull their weight, with no 1-line wrappers between them.

**Hex decoder.** Folded `ManifestHex.kt` back into `ManifestVerifier.kt` and replaced the hand-rolled decoder with `java.util.HexFormat` (JVM 17 is the toolchain target — `KotlinLibraryConventionPlugin` sets `JvmTarget.JVM_17`). The explicit length check stays because `HexFormat.parseHex` accepts any even length and we require exactly SHA-1.

`:passes-core:check` and `:passes-core:detekt` green; `PublicApiSurfaceTest` unchanged. 23 `ManifestVerifierTest` cases pass.

PR #26, opened 2026-05-05
passes-core: pass.json deserialization slice (wpass-7sl)

bittelc:
## Summary

Lands the internal `decodePassJson(entries, config)` slice — the second of three `PassParser` implementation slices, after `wpass-8so` (extractor) and `wpass-7l2` (manifest hash chain).

- Pre-pass tokenizer enforces `ParserConfig.maxJsonDepth` and `maxJsonStringBytes` *before* kotlinx.serialization sees the bytes. kotlinx does not enforce these natively, and post-validating the parsed tree is too late — a deeply-nested adversarial payload has already allocated by the time the tree exists.
- Typed parse runs in strict mode (`isLenient = false`) with `ignoreUnknownKeys = true` so PKPASS spec additions do not break consumers.
- Internal failure arms: `Missing`, `InvalidJson`, `InvalidShape`, `JsonDepthExceeded`, `JsonStringTooLong`, `UnknownFormatVersion(v)`, `UnknownPassStyle(raw)`. Mapping to the public `ParseResult` lands in `wpass-qnc`; this slice keeps the public API surface frozen (`PublicApiSurfaceTest` unchanged).
- Dangerous fields (`nfc`, `webServiceURL`, `authenticationToken`, `personalization`, `personalizationToken`) are structurally absent from the `Pass` model per `decision-wlt-0tn-q1` — kotlinx parses them under `ignoreUnknownKeys=true` and the data shape has nowhere to surface them, keeping the trust claim auditable from the model itself rather than via runtime stripping.
- Returned `Pass` carries `images = emptyMap()` and `locales = emptyMap()`; the parser-glue bead wires those in.

## Test plan

- [x] `./gradlew :passes-core:check :passes-core:detekt` — all green
- [x] 30 new `PassJsonDecoderTest` cases: each `PassType`, missing top-level fields, unknown `formatVersion`, unknown style key, `rgb(...)` vs `#RRGGBB`, `barcode` vs `barcodes[0]`, dangerous fields parsed-and-dropped (asserted via structural equality with the canonical pass), `maxJsonDepth` and `maxJsonStringBytes` trips, escape-aware string-length counter, per-string counter resets, boundary-at-the-limit acceptance.
- [x] `PublicApiSurfaceTest` unchanged — verified at the file level (no diff) and as a passing test.

Closes once main CI is green: `wpass-7sl`.

### DIFF
--- a/passes-core/src/main/kotlin/is/walt/passes/core/internal/JsonLimitTokenizer.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/internal/JsonLimitTokenizer.kt
@@ -1,0 +1,98 @@
+package `is`.walt.passes.core.internal
+
+import `is`.walt.passes.core.ParserConfig
+
+/**
+ * Defensive ceiling check the kotlinx parser does not natively enforce. Iterates
+ * source bytes once, tracking nesting depth and the byte length of in-progress JSON
+ * string tokens. ASCII-byte scanning is safe over UTF-8: continuation bytes
+ * (`0x80..0xBF`) cannot collide with `{`, `}`, `[`, `]`, `"`, or `\`.
+ *
+ * String byte-counting uses source bytes (overcount by escape-sequence shrinkage),
+ * not decoded character bytes — the guard's intent is to bound JSON-bomb expansion
+ * before allocation, and the overcount is conservative (never lets an over-budget
+ * string through). Returns `null` on success, or the first arm that tripped. JSON
+ * well-formedness is intentionally not verified here — kotlinx.serialization handles
+ * that downstream — so an unbalanced bracket pair sails through here and surfaces as
+ * [PassJsonFailure.InvalidJson] from the typed parse.
+ */
+internal fun enforceJsonLimits(
+    bytes: ByteArray,
+    config: ParserConfig,
+): PassJsonFailure? {
+    val state =
+        JsonLimitTokenizer(
+            maxDepth = config.maxJsonDepth,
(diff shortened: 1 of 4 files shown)


bittelc:
Thanks for the review. Addressed in b07be82:

- **Depth clamp** added to `JsonLimitTokenizer.consumeOutsideString`. Stray leading closers no longer drive `depth` negative; the invariant `0 <= depth <= maxDepth` now holds without relying on kotlinx or the entry-size cap to bound the peak. Regression test `strayLeadingClosersDoNotInflateInFlightDepth` exercises a `}}{{{}}` payload at `maxDepth=2`, which would surface as InvalidJson without the clamp and now fires JsonDepthExceeded at the third opener.
- **Tokenizer extracted** to `JsonLimitTokenizer.kt`. PassJsonDecoder.kt is now purely the structural-mapping layer.
- **Routing-table doc duplication trimmed.** Per-arm KDoc on `PassJsonFailure` is one line each; the canonical defense-in-depth statement lives in `PassJsonDecoder.kt`'s file header.
- **Apology comment removed** from `assemblePass`. Restructured around a `RequiredFields` extraction; the smart-cast / detekt-complexity meta-comment is gone.

Filed the two future-cleanup caveats:

- `wpass-k5t` (P3): relax expirationDate parsing for naked-local timestamps if real-world fixtures show this fires.
- `wpass-rj7` (P4): nullable `version` on `UnknownFormatVersion` to distinguish missing from non-1 without a sentinel.

Quality gates green locally; CI re-running.

PR #29, opened 2026-05-06
passes-core: PassParser.create() factory glue (wpass-qnc)

bittelc:
## Summary

- Replaces the `NotImplementedError` stub in `PassParser.create()` with `DefaultPassParser`, wiring the four merged implementation slices (extract, manifest, signature, pass.json) plus `.strings` parsing into a non-throwing end-to-end pipeline.
- Adds a pure-JVM IHDR-only PNG header reader so `maxImagePixelCount` fires before any decoder allocates pixels.
- Brackets every `parse()` with `TelemetryGuard` events carrying `nanoTime`-derived durations and type-flattened status / failure kinds; pass content never enters the events.

Closes `wpass-qnc`. Unblocks walt-android `wlt-4pg` (consumer module-graph wiring).

### Pipeline order (trust hierarchy)

```
extract -> manifest hash chain -> CMS signature -> pass.json -> .strings -> images -> assemble
```

Each stage's failure lands on its public arm: `Tampered` for hash mismatch and signature failures, `Malformed` for structural issues, `Unsupported` for unknown `formatVersion` / pass style, `ResourceLimitExceeded` for guard trips.

### Concurrency

`DefaultPassParser` holds only the immutable `ParserConfig`; helpers take work-in-progress entries as parameters. Safe to call `parse()` concurrently across threads on a single instance.

## Test plan

- [x] `./gradlew :passes-core:check` green (detekt + ktlint + JUnit, 169 tests)
- [x] All five `PassType` happy paths produce `Success`
- [x] Unsigned + lenient default -> `Success(SignatureStatus.Unsigned)`
- [x] Unsigned + `ParserConfig.Strict` -> `Tampered(SignatureCryptoFailure)`
- [x] Self-signed + lenient -> `Success(SignatureStatus.SelfSigned)`; strict rejects
- [x] One-byte tampered `pass.json` -> `Tampered(FileHashMismatch)`
- [x] `Malformed`/`Unsupported` arms forwarded for missing/invalid pass.json, unknown format version, non-zip bytes
- [x] Resource caps trip: `EntrySize` (zip-bomb), `ImagePixelCount` (oversized IHDR), `LocaleCount` (more `.lproj/` than the cap)
- [x] Locales and image roles surfaced on the parsed `Pass`
- [x] Telemetry: `onParseStarted` followed by exactly one of `onParseSucceeded` / `onParseFailed`
- [x] 16-thread concurrent parse on a shared parser + source returns equal `Success`
- [x] `PublicApiSurfaceTest.parserCanParseSyntheticPkpass` locks the no-throw end-to-end surface

### DIFF
--- a/passes-core/src/main/kotlin/is/walt/passes/core/PassParser.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/PassParser.kt
@@ -1,5 +1,6 @@
 package `is`.walt.passes.core
 
+import `is`.walt.passes.core.internal.DefaultPassParser
 import java.io.InputStream
 
 /**
@@ -13,20 +14,25 @@ import java.io.InputStream
  * UI must distinguish.
  */
 public fun interface PassParser {
+    /**
+     * Parse [source] into a [ParseResult]. Synchronous and CPU-bound: BouncyCastle's
+     * PKCS#7 verification, JSON tokenization, and zip extraction all run on the calling
+     * thread. Wrap with `withTimeout(...)` (kotlinx.coroutines) when invoking from a
+     * coroutine context that requires a per-call timeout — [ParserConfig] does not yet
+     * expose a built-in deadline knob and will not until the public surface freeze
+     * around ADR 0001 lifts.
+     *
+     * Never throws: every failure mode is encoded as a [ParseResult] arm.
+     */
     public fun parse(source: PassSource): ParseResult
 
     public companion object {
         /**
-         * Construct a parser bound to [config]. The default factory is intentionally not
-         * yet implemented; landing it is the next bead after this skeleton.
+         * Construct a parser bound to [config]. The returned instance is safe to call
+         * concurrently from multiple threads — it holds only the immutable [config] and
+         * delegates to stateless internal pipeline functions on each [parse] invocation.
          */
-        public fun create(config: ParserConfig = ParserConfig()): PassParser =
-            PassParser { _ ->
-                throw NotImplementedError(
-                    "PassParser implementation pending follow-up bead; this skeleton only " +
-                        "fixes the public API surface. config=$config",
-                )
-            }
+        public fun create(config: ParserConfig = ParserConfig()): PassParser = DefaultPassParser(config)
     }
 }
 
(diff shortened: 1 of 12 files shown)


bittelc:
Thanks — all five concrete issues addressed in d7128da.

1. **PNG pixel-count overflow** — pre-check each IHDR axis against the cap before multiplying. If either axis already exceeds the cap, we trip without doing the multiplication; otherwise both axes are ≤ Int.MAX_VALUE and the product is safely under Long.MAX_VALUE. New test `pngPixelCountLimitTripsForU32MaxIhdrWithoutOverflowing` exercises `width = height = Int.MAX_VALUE`.

2. **Flatten the helper chain** — collapsed `withEntries`/`withVerifiedManifest`/`withSignatureStatus`/`assemble` into a single `runPipeline` reading top-to-bottom with `@Suppress("ReturnCount")` and a comment explaining why (the chain-of-helpers existed only to satisfy detekt's `ReturnCount: 2`; you're right that it hides the linear shape).

3. **Three phase types collapsed** to one `Phase<T> = Continue<T> | Halt(ParseResult)`.

4. **archiveBytes for hint-less streams** — `BoundedInputStream` now exposes its measured count, plumbed through `ExtractResult.Success.archiveBytes`. `computeArchiveBytes` uses source-derived size when known (`PassSource.Bytes`, hint-bearing `Stream`) and the extractor's count as fallback. New test `streamSourceWithoutSizeHintReportsExtractedBytesOnTelemetry`.

5. **Shared signature/manifest constants** — hoisted to `internal/Constants.kt` (`SIGNATURE_FILE_NAME`, `MANIFEST_FILE_NAME`); `SafeArchiveExtractor`, `ManifestVerifier`, and `DefaultPassParser` now share the same source.

Bonus thread (test fixture dedup): extracted `newRsaKeyPair` / `selfSignedCertificate` / `signLeafCertificate` / `cmsDetachedSignature` into `internal/TestCryptoSupport.kt`. `SignatureVerifierTest` and `SyntheticPkpass` use the same helpers; the duplicated CMS signing block is gone.

Test count: 169 → 171 (PassParserTest 23 → 25). All green: detekt, ktlint, junit.

PR #36, opened 2026-05-06
passes-pdf: isolated-process renderer service (wpass-5v9)

bittelc:
## Summary
- New `passes-pdf` Android module: isolated-process `PdfRendererService` wrapping Android's PdfRenderer (PDFium). Manifest declares **zero** `uses-permission` entries; `android:isolatedProcess="true"`, `android:exported="false"`.
- Internal binder is a hand-rolled `Binder` subclass (no AIDL). Surface is exactly `probe` + `render` — no `getText`, `getMetadata`, `getAnnotations`, `getAttachments`, or `getFormFields`. `PublicApiSurfaceTest` reflection-locks the surface.
- Render watchdog enforces `PdfImportConfig.renderTimeoutMs` (5 s default) and kills the renderer process on timeout via `Process.killProcess`. Killer is injected via `ProcessKiller` so the unit test exercises the kill path without taking down the test JVM.
- Caps: `MAX_BYTES = 25 MB`, `MAX_PAGES = 10`, `MAX_PIXELS = 4 MP` — mirrored from `passes-storage` `DocumentBounds`. Cross-module parity test tracked separately in `wpass-kej`; until then both numbers are kept in lockstep manually with reciprocal comments.
- `ManifestPermissionsTest` parses the source manifest and asserts zero `uses-permission` entries plus `isolatedProcess="true"` + `exported="false"`. Pure-JVM so it runs under `./gradlew check`.
- Instrumented coverage (benign / JS-laden / malformed / encrypted / 11-page) is checked in but `@Ignore`'d pending fixture set + on-device CI wiring.

## Test plan
- [x] `./gradlew check` — green (full repo)
- [x] `./gradlew :passes-pdf:testDebugUnitTest` — `RenderWatchdogTest`, `PublicApiSurfaceTest`, `ManifestPermissionsTest` pass
- [ ] On-device: `./gradlew :passes-pdf:connectedDebugAndroidTest` after fixture set lands (tracked by `wpass-5v9` follow-up)
- [ ] CI lint asserting zero `uses-permission` on the merged consumer manifest (out of scope for this PR; should land alongside the consumer wiring in `wlt-4pg`)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,12 +10,14 @@ The trust claim: every security-and-privacy-critical behavior Walt makes about p
 
 ## Architecture Rules
 
-- Three Gradle modules:
+- Five Gradle modules:
   - `passes-core` — pure Kotlin/JVM. Parser, model, signature verifier, `.strings` parser, `TelemetryGuard` interface. NO Android framework dependencies.
+  - `passes-pdf-core` — pure Kotlin/JVM. PdfDocument model, header sniffer, import config, `DocumentTelemetryGuard` interface. NO Android framework dependencies.
+  - `passes-pdf` — Android-only. Isolated-process PDF renderer service (ADR 0005 D3). Wraps Android's `PdfRenderer` behind a hand-rolled binder with no extraction surface.
   - `passes-storage` — Android-only. SQLCipher with Keystore-sourced key, Android Auto Backup exclusion, irreversible deletion logic.
   - `passes-ui` — Android + Compose. Pass front/back composables, B3 URL confirmation sheet, expired badge, bounded image rendering. Themable via tokens passed in by the consumer.
-- `passes-core` has NO Android framework dependencies (KMP-friendly).
-- `passes-storage` and `passes-ui` are independent of each other.
+- `passes-core` and `passes-pdf-core` have NO Android framework dependencies (KMP-friendly).
+- `passes-storage`, `passes-pdf`, and `passes-ui` are independent of each other.
 - DECISIVE CONSTRAINT: walt-android consumes this code directly; trust-claim-bearing logic lives ONLY here, never reimplemented in walt-android.
 
 ## Code Style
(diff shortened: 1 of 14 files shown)


bittelc:
Pushed [a96dede](https://github.com/walt-app/walt-passes-android/pull/36/commits/a96dede) addressing the review:

**1. Watchdog kill-path now fires for uncancellable JNI hangs.**
- `RenderWatchdog.guard` now runs the timer as a sibling of the guarded work via `coroutineScope { launch killer; try { withContext(IO) { block() } } finally { killerJob.cancel() } }`. The killer's own `delay` runs on the coroutine framework's timer and is independent of whether the guarded block reaches a suspension point.
- New test `timeoutFiresKillerForUncancellableNativeShapedBlock` uses a `Thread.sleep`-based polling block that stands in for a stuck PDFium frame; the test passes (61ms wall-clock) and would hang under the old `withTimeout`-only design — JUnit `@Test(timeout = 3000)` is the regression-net.

**3. `doRender` enforces MAX_PAGES.** A 5000-page PDF cannot be rasterised even if a caller skips probe; both `renderer.pageCount > MAX_PAGES` and `page >= MAX_PAGES` now reject as `RendererFailed` before `openPage` runs.

**5. `CLAUDE.md` refreshed.** Module list now lists all five (`passes-core`, `passes-pdf-core`, `passes-pdf`, `passes-storage`, `passes-ui`); the "no Android framework deps" rule is now pinned to `passes-pdf-core` too.

**2. Client-side binder proxy + round-trip test → `wpass-11u`.** Filed as a P1 follow-up that blocks `wpass-7wn` (passes-ui DocumentView) and the walt-android consumer wiring in `wlt-4pg`. The bead also calls out replacing `DocumentRejectedKind.ordinal` with an explicit stable wire mapping so an enum reorder is a structural failure rather than silent wire-compat drift.

**4. Bitmap sequencing.** Left as-is per your "not a defect" note.
**6. ManifestPermissionsTest cwd nit.** Left as-is — Gradle sets cwd reliably; happy to revisit if/when we see IDE-runner pain.

`./gradlew check` green; the new `RenderWatchdogTest` runs in ~170ms total.

PR #38, opened 2026-05-06
passes-ui: DocumentView composable + Documents lane + trust caption (wpass-7wn)

bittelc:
## Summary

Composable surfaces for PDF documents: `DocumentTrustCaption` (non-suppressible), `DocumentTile`, `DocumentsLane`, and `DocumentView` with `HorizontalPager` driving an LRU-bounded bitmap cache through the `PdfRendererBinder` *interface* (not the concrete client) so tests inject fakes without binding the isolated-process service.

Closes wpass-7wn.

## Trust-claim discipline

Mirrors `ExpiredOverlay`: no `enabled` flag, no theme token that hides the caption, no `DocumentView` overload that skips it.

- `ComposableSurfaceLockTest` pins parameter shapes for the four new composables and asserts none has an overload (a future `DocumentView(..., showCaption: Boolean)` sibling fails the test).
- `TrustClaimSurfaceTest` asserts the caption renders the verbatim audit text and that `displayLabel` is BiDi-isolated with U+2068/U+2069 on the tile (mirrors the B3 sheet pattern in `SecuritySheets.kt`).
- `PublicApiSurfaceTest` gains a bytecode scan that walks every `is.walt.passes.ui.*` (and reachable `is.walt.passes.pdf.*`) compiled class and fails closed if it finds the constants `android.intent.action.SEND`, `android.intent.action.SEND_MULTIPLE`, or `application/pdf`. ADR 0005 D8 (no share-out) and D4 (no extraction) become structural locks rather than documentation.

## LRU cache

`RenderedPageCache<V>` is generic over the value type so the access-order eviction contract is exercisable JVM-pure (eight test cases in `RenderedPageCacheTest`). Production wiring binds `Bitmap` with `Bitmap::recycle` as the eviction handler so native pixel memory is freed the moment a page falls out of the 3-page window. The renderer service has already pre-recycled its source-side copy (per ADR 0005 D3); the UI is the sole bitmap owner past the binder.

## Module-graph caveat — flagged for review

`passes-ui` now declares `api(project(":passes-pdf"))`, which crosses the rule documented in `CLAUDE.md`: *"passes-storage, passes-pdf, and passes-ui are independent of each other."*

The alternative would be a new `passes-pdf-ui` module, splitting `DocumentView` from `PassFront` across modules even though both are pass-style trust-claim surfaces sharing the same theming and BiDi-isolation utilities. The dependency is taken in this PR with a flagged inline comment in `passes-ui/build.gradle.kts`; reviewers should decide whether to bend the architectural rule or split the module before merge.

## Theming

`DocumentSemantics` is a sibling data class to the other slot families, exposed as `PassesSemantics.documents`. Walt-android's `core/ui` is authoritative for the actual hex values (per `CLAUDE.local.md`); `passes-ui` owns the contract shape only.

## Test plan

- [x] `./gradlew check` — passes-core, passes-pdf-core, passes-pdf, passes-storage, passes-ui all green; ktlint, detekt, lint, all unit tests
- [x] passes-ui unit tests: `RenderedPageCacheTest` (8), `TrustClaimSurfaceTest` (18, including new doc-surface and BiDi-isolation tests), `ComposableSurfaceLockTest` (15, including new param-count locks for the four new composables and a no-overload assertion), `PublicApiSurfaceTest` (17, including the new bytecode-scan tests over passes-ui and passes-pdf compiled classes)
- [ ] Instrumented test (`DocumentViewInstrumentedTest`): pager-driven render(), caption visibility under render rejection, distinct render() calls past the LRU window. Requires an emulator/device run that this PR's CI does not yet do — host walt-android is the canonical site for that run; this branch ships the test for future invocation.
- [ ] Visual review on emulator with both light and dark schemes from walt-android's `core/ui` (host theming) — host-side validation, not part of this branch.

### DIFF
--- a/passes-ui/build.gradle.kts
+++ b/passes-ui/build.gradle.kts
@@ -11,6 +11,18 @@ android {
 dependencies {
     api(project(":passes-core"))
 
+    // ARCHITECTURAL CAVEAT (CLAUDE.md "passes-storage, passes-pdf, and passes-ui are
+    // independent of each other"): adding passes-pdf here crosses that rule. The
+    // alternative — a separate passes-pdf-ui module — would split DocumentView from
+    // PassFront across modules even though both are pass-style trust-claim surfaces
+    // sharing the same theming and isolation utilities. We accept the dependency so
+    // composables that present documents live next to composables that present passes;
+    // see the PR description for the trade-off discussion. The dep is `api` because
+    // PdfRendererBinder is a parameter type on DocumentView (callers must construct
+    // one), so pulling it in transitively is the usable shape for consumers.
+    api(project(":passes-pdf"))
+    api(libs.kotlinx.coroutines.core)
+
     implementation(libs.androidx.core.ktx)
     implementation(libs.androidx.lifecycle.runtime.ktx)
 
@@ -38,4 +50,5 @@ dependencies {
     androidTestImplementation(composeBom)
     androidTestImplementation(libs.androidx.test.junit)
     androidTestImplementation(libs.androidx.compose.ui.test.junit4)
+    androidTestImplementation(libs.truth)
 }
--- a/passes-ui/src/androidTest/kotlin/is/walt/passes/ui/DocumentViewInstrumentedTest.kt
+++ b/passes-ui/src/androidTest/kotlin/is/walt/passes/ui/DocumentViewInstrumentedTest.kt
@@ -1,0 +1,260 @@
+package `is`.walt.passes.ui
+
+import android.os.ParcelFileDescriptor
+import android.os.SharedMemory
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.onRoot
+import androidx.compose.ui.test.performTouchInput
+import androidx.compose.ui.test.swipeLeft
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.google.common.truth.Truth.assertThat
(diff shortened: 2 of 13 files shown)


bittelc:
Addressed in 6d22706 (just pushed).

**In this PR (block-on):**
- C1: `bitmapFromSharedMemory` now uses `ok.widthPx`/`ok.heightPx` from `RenderResult.Ok`, not the caller's request. Real correctness bug — using request dimensions against a renderer-downscaled `SharedMemory` was throwing inside `copyPixelsFromBuffer` and getting swallowed by `runCatching`.
- C4: `Image(contentDescription = "Page N of M")` for TalkBack. Positional caption is the only safe fallback under D4.
- C2: `LRU_PAGE_WINDOW` 3 → 5 (current ± 2). Comment in source explains the HorizontalPager retention reasoning.
- A2: `DocumentView` KDoc now states the `pdfFile` lifetime contract.
- A3: dropped the dead `pageCount == 0` branch.
- C7: removed redundant inner `aspectRatio`.
- Added a maintainability comment at `TRUST_CAPTION_TEXT` documenting the deliberate dual-anchor placement so a future contributor doesn't "DRY" it.

**Follow-ups filed:**
- `wpass-r4z` — extract `passes-pdf-ui` or split utilities so `passes-ui` no longer depends on `passes-pdf` via `api`. Restores module independence (your A1).
- `wpass-8v4` — `DocumentTelemetryGuard` arm for consumer-side render failures. The `runCatching` swallow path now carries a justifying comment pointing at this issue (your C5).
- `wpass-8rx` — sentinel pixel pattern in instrumented-test fakes so a C1-class regression is catchable on device (your T1).

`./gradlew :passes-ui:check` green; `:passes-ui:compileDebugAndroidTestKotlin` green.

T2 / T3 / dual-anchor comment: noted, no action besides the comment at the constant.

PR #50, opened 2026-05-07
passes-core: hold BouncyCastle instance, not "BC" name lookup (wpass-4js)

bittelc:
## Summary

- **Root cause**: `ensureBouncyCastleRegistered` checked `Security.getProvider("BC") == null`, which is always false on Android (it ships a stripped-down BC under the `"BC"` slot). So `Security.addProvider(BouncyCastleProvider())` never fired, and subsequent `setProvider("BC")` calls resolved to Android's stripped instance — which mishandles CMS / PKCS#7 verification and surfaced every Apple-signed pkpass as `ParseResult.Tampered`.
- **Fix**: `SignatureVerifier` now holds its own `BouncyCastleProvider` instance and passes it directly to `JcaSimpleSignerInfoVerifierBuilder`, `JcaX509CertificateConverter`, and `CertPathBuilder` via their `Provider`-instance overloads. JCE name-registry ordering is no longer load-bearing.
- **Hardening**: new `TamperReason.SignerCertificateMissing` arm distinguishes "envelope parses but signer cert absent in cert set" from generic crypto failure, so a future regression on this code path stays visible in telemetry instead of collapsing onto the catch-all.

The bd's hypothesis (that the SKI signer-ID path itself was broken) was not the cause: a synthetic SKI test passes against the verifier on JVM unmodified. The combination of (a) Android-stripped BC occupying the `"BC"` slot and (b) name-lookup-based provider resolution is what closed the loop.

## Test plan

- [x] `:passes-core:test` (8 suites, 176 tests) passes
- [x] `:passes-core:detekt` clean
- [x] New regression test `realAppleSignedPkpassVerifiesEvenWhenBcSlotHoldsAStrippedProvider` parks a non-BC provider in the `"BC"` slot and confirms verification still succeeds. Verified the test FAILS without the fix (returns `Failed(SignatureCryptoFailure)` — exactly the field-observed shape).
- [x] New synthetic SKI signer-ID tests (`appleVerifiedWhenSignerIdIsSubjectKeyIdentifier`, `manifestSignatureMismatchForSkiSignerWithTamperedManifest`) lock the Apple-realistic signer-ID flavor in place.
- [x] New real-Apple-signed fixture test (`realAppleSignedPkpassFixtureVerifies`) consumes `manifest.json` + `signature` extracted from a Tixly-signed pkpass and asserts `AppleVerified` end-to-end.
- [x] New `signerCertificateMissingWhenSignerIdMatchesNoCertInEnvelope` exercises the new `TamperReason` arm.

## Files

- `passes-core/src/main/kotlin/is/walt/passes/core/ParseResult.kt` — new `TamperReason.SignerCertificateMissing` arm
- `passes-core/src/main/kotlin/is/walt/passes/core/internal/SignatureVerifier.kt` — held-instance refactor, removed `ensureBouncyCastleRegistered`, route the `firstSignerWithCert == null` path to the new arm
- `passes-core/src/test/.../TestCryptoSupport.kt` — `cmsDetachedSignatureWithSki` helper, certs now carry SubjectKeyIdentifier extensions by default
- `passes-core/src/test/.../SignatureVerifierTest.kt` — five new tests (registry-poisoning, real-fixture, two SKI-synthetic, one signer-cert-missing)
- `passes-core/src/test/.../PublicApiSurfaceTest.kt` — extend the TamperReason exhaustiveness check
- `passes-core/src/test/resources/is/walt/passes/core/internal/fixtures/apple-signed/{manifest.json, signature}` — real Apple-signed fixture (manifest is just file digests, signature embeds the WWDR-G4 + Tixly leaf certs; no user PII)

Closes wpass-4js.


bittelc:
On-device probe (BcProviderProbeTest, run on FP4 / Android 16) confirms the root cause this PR fixes:

- `Security.getProvider("BC")` → AOSP-stripped BC 1.77 (`com.android.org.bouncycastle...`), shadowing our bundled bcprov-jdk18on 1.79.
- `Signature.getInstance("SHA256withRSA", "BC")` throws `NoSuchAlgorithmException` — the AOSP fork has 105 services, none of them `Signature.SHA256withRSA`.
- That exception is what the verifier's outer `runCatching` was absorbing, surfacing every Apple-signed pkpass as `Failed(SignatureCryptoFailure)`. Matches the field telemetry exactly.
- `ensureBouncyCastleRegistered`'s `if (Security.getProvider("BC") == null)` is always false on Android, so the unbundled BC was never installed — confirmed.

This makes the **SKI signer-ID hypothesis a red herring** for this bug: failure fires for every Apple-signed pkpass on Android 9+ regardless of signer-ID flavor. The synthetic SKI tests added here (`appleVerifiedWhenSignerIdIsSubjectKeyIdentifier`, `manifestSignatureMismatchForSkiSignerWithTamperedManifest`) stay in the suite as forward coverage for the SKI path — but the wpass-4js *regression* test is `realAppleSignedPkpassVerifiesEvenWhenBcSlotHoldsAStrippedProvider`, which simulates the AOSP shape via a name-collision in the `"BC"` slot and reproduces the exact field-observed `Failed(SignatureCryptoFailure)` shape pre-fix. Test comments and the `BC_PROVIDER` docstring updated in [f67b48f](https://github.com/walt-app/walt-passes-android/commit/f67b48f) to match the confirmed mechanism.

**wpass-aio (SQLCipher DatabaseCorrupt) — does not share this root cause.** Scanned `passes-storage`: zero `"BC"` strings, zero `BouncyCastle` references. The SQLCipher path is `net.zetetic.database.sqlcipher` (native crypto, libsqlcipher.so — not JCE), and `AndroidKeystorePassKeyProvider`'s `Cipher.getInstance("AES/GCM/NoPadding")` calls have no provider arg and rely on AndroidKeyStore-bound key auto-routing. wpass-aio needs its own diagnosis.

**Re: BcProviderProbeTest in the wlt-a07 worktree** — `passes-core` is pure Kotlin/JVM (no Android framework dependency, no `androidTest` source set), so the probe can't be lifted here. The natural home is `passes-storage`'s `androidTest` (Android-only module), or kept in walt-android as the closed-source consumer-side smoke test. Happy to file a follow-up to add it to `passes-storage` if useful — it'd be a single-class device-truth check that documents the AOSP BC shape we just dodged.

bittelc:
Thanks for the review — addressed in [c0e2f0d](https://github.com/walt-app/walt-passes-android/commit/c0e2f0d):

1. **SignerCertificateMissing doc accuracy** — docstring now covers both reaching shapes (zero-SignerInfo envelopes and signer-with-no-matching-cert), and explains why folding them together is intentional: the operational signal is identical (well-formed but unsignable, which a corrupt blob is not). Restructuring `firstSignerWithCert` into a richer sum type was the cleaner alternative the review offered, but since both shapes deserve the same telemetry routing in production, the docstring approach is the simpler match.

2. **Cross-file `BC_PROVIDER` shadowing** — renamed the test-side constant to `BC_PROVIDER_NAME`. Added a comment at its declaration explaining why the test path can keep using string lookup (JVM-only, our `ensureBouncyCastleProvider` registers BC 1.79 cleanly without an AOSP shadow). The `_NAME` suffix flags the type difference for any reader skimming both files.

3. **Concurrency contract on the registry-mutating test** — the regression test now carries a "Concurrency contract" paragraph documenting that JUnit 4 + Gradle default `maxParallelForks=1` makes this safe today, and what to do (`@Isolated`, dedicated single-threaded fork) if anyone changes that.

4. **Fixture shelf life** — added `passes-core/src/test/resources/.../fixtures/apple-signed/README.md`. Records the current Tixly leaf's `notAfter` (`2027-02-04T10:18:41Z`), enumerates what is *not* in the fixture (`pass.json`, `*.lproj/`, images — the PII-bearing entries), and gives a step-by-step renewal procedure for the inevitable "test broke and nobody knows why" moment. The renewal is "grab any current Apple-signed pkpass, copy its `manifest.json` + `signature` over." `realAppleSignedPkpassFixtureVerifies` now points at the README.

5. **Single-signer scope on `firstSignerWithCert`** — added a docblock making the constraint explicit ("Apple emits exactly one SignerInfo; multi-signer CMS is out of scope by design") and noting that this is the entry point a future bead would need to fan out if the trust claim ever extends to multi-signer envelopes.

6. **`Pair<ByteArray, ByteArray>` → named data class** — fixed. `loadAppleSignedFixture()` now returns `AppleSignedFixture(manifest, signature)`. The two callers updated; both ByteArrays at the call site are now field-named so they can't be transposed into the wrong slot.

**Deferred (review item 5):** `setDate()` on `PKIXBuilderParameters` snapshotted to the CMS signing time. Already flagged in `SignatureVerifier.kt`'s "No date-shift override" docblock. Will file a follow-up bead so the renewal-procedure churn for the real fixture has a path to retirement.

176 tests + detekt green.

PR #51, opened 2026-05-07
passes-storage: keep DatabaseKey buffer alive for the SQLCipher pool (wpass-aio)

bittelc:
## Summary

Fixes wpass-aio. On fresh waltDebug install, the Passes screen surfaces "Secure storage is unavailable" with `passes_storage_failure[kind=Unknown, unknown_kind=DatabaseCorrupt]`. With cc41451's diagnostic logging in place, the underlying throw is `SQLiteOutOfMemoryException` from `nativeExecuteForCursorWindow` at the schema-probe `SELECT 1 FROM sqlite_master ...` - the first cursor-backed read after `openOrCreateDatabase`.

### Root cause (verified against `net.zetetic` 4.6.1 source)

- [`SQLiteDatabaseConfiguration.java:93,134`](https://github.com/sqlcipher/sqlcipher-android/blob/master/sqlcipher/src/main/java/net/zetetic/database/sqlcipher/SQLiteDatabaseConfiguration.java) holds the `password` byte[] **by reference** (`this.password = password`, no copy).
- [`SQLiteConnection.open()`](https://github.com/sqlcipher/sqlcipher-android/blob/master/sqlcipher/src/main/java/net/zetetic/database/sqlcipher/SQLiteConnection.java) re-reads `mConfiguration.password` and calls `nativeKey()` on **every** pool connection it opens - including read-only auxiliaries opened lazily on the first cursor read.

The previous `databaseKey.withBytes { ... openOrCreateDatabase ... }` lifetime zeroed the buffer when the lambda returned. The primary connection got keyed correctly during the open call (so file lands at 61440 bytes, PRAGMA execSQL succeeds), but the first `rawQuery` acquires a fresh pool connection that re-keys with all-zero bytes. Page-1 decrypt fails inside the validation `SELECT COUNT(*) FROM sqlite_schema` and surfaces as the observed `SQLITE_NOMEM`.

### Fix

Make the byte[] outlive the connection pool:

- `DatabaseKey.retainAcross(block) -> AutoCloseable` hands the bytes to a long-lived consumer and returns a handle whose `close()` zeros the buffer. `withBytes` is unchanged - it still zeros synchronously for short-lived borrows (KDFs, JVM-test helpers).
- `SqlCipherDatabaseFactory.openOrCreate` returns `OpenedDatabase(db, keyHandle)` and threads the keyHandle through every failure path so the buffer is zeroed even on setup-phase aborts.
- `SqlCipherPassStore.close()` closes the SQLiteDatabase first, then the keyHandle - reversing the order would zero the buffer while the pool may still be re-keying connections during shutdown.

### Why not the alternative diagnoses

- **PRAGMA cipher_compatibility ordering**: `sqlcipher-android` 4.6.1's default is already compat=4, so the user-issued `PRAGMA cipher_compatibility = 4` is a redundant no-op, not a regime change. No on-disk-vs-runtime mismatch on a fresh install. (Worth a follow-up to drop the redundant PRAGMA, but not the fix for this bug.)
- **`SQLiteDatabaseHook` + `postKey`**: reorders PRAGMAs within a single connection's `open()` but doesn't change which byte[] `nativeKey` reads. Pool-acquired auxiliaries would still re-key with zeros.
- **Move PRAGMA + DDL inside `withBytes` (option 1)**: passes startup but leaves a latent bug. Any future pool connection acquired after setup (concurrent reads, post-resume reconnections) re-keys with zeros and fails the same way.

## Test plan

- [x] `./gradlew :passes-storage:compileDebugKotlin :passes-storage:compileReleaseKotlin`
- [x] `./gradlew :passes-storage:testDebugUnitTest` (includes new `databaseKeyRetainAcrossKeepsBytesAliveUntilHandleClosed` pinning the lifetime contract)
- [x] `./gradlew :passes-storage:detekt` (baseline updated for the changed `openOrCreate` return signature)
- [x] `./gradlew :passes-storage:assembleDebug :passes-storage:assembleRelease`
- [ ] Smoke test on FP4 / Android 16 waltDebug: fresh install -> open Passes screen -> no `passes_storage_failure`; import a pass -> reopen the app -> existing pass renders. (Needs walt-android consumer wiring; will be verified during wlt-4pg integration.)

### DIFF
--- a/passes-storage/detekt-baseline.xml
+++ b/passes-storage/detekt-baseline.xml
@@ -9,7 +9,7 @@
     <ID>NestedBlockDepth:SqlCipherPassStore.kt$SqlCipherPassStore$override fun upsert( pass: Pass, signatureStatus: SignatureStatus, nowEpochMs: Long, ): UpsertOutcome</ID>
     <ID>ReturnCount:AndroidKeystorePassKeyProvider.kt$AndroidKeystorePassKeyProvider$override fun provideDatabaseKey(): StorageResult&lt;DatabaseKey&gt;</ID>
     <ID>ReturnCount:BackupRulesAssertion.kt$BackupRulesAssertion$@JvmStatic public fun assertBackupRulesApplied(context: Context): Outcome</ID>
-    <ID>ReturnCount:SqlCipherDatabaseFactory.kt$SqlCipherDatabaseFactory$fun openOrCreate( context: Context, databaseKey: DatabaseKey, ): StorageResult&lt;SQLiteDatabase&gt;</ID>
+    <ID>ReturnCount:SqlCipherDatabaseFactory.kt$SqlCipherDatabaseFactory$fun openOrCreate( context: Context, databaseKey: DatabaseKey, ): StorageResult&lt;OpenedDatabase&gt;</ID>
     <ID>ReturnCount:SqlCipherPassRepository.kt$SqlCipherPassRepository.Companion$@JvmStatic public fun create( context: Context, keyProvider: PassKeyProvider, telemetryGuard: StorageTelemetryGuard, ioDispatcher: CoroutineDispatcher = Dispatchers.IO, clock: () -&gt; Long = { System.currentTimeMillis() }, ): StorageResult&lt;SqlCipherPassRepository&gt;</ID>
     <ID>ReturnCount:SqlCipherPassStore.kt$SqlCipherPassStore$override fun loadById(id: PassRecordId): StoredPass?</ID>
     <ID>ReturnCount:SqlCipherPassStore.kt$SqlCipherPassStore$private fun Cursor.toSummaryOrNull(): PassSummary?</ID>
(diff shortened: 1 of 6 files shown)


bittelc:
Addresses the second review's #1 (real correctness gap: `retainAcross` doesn't zero on exception).

**Pushed in 736604d:** `retainAcross` now wraps the block in `try { ... } catch (t: Throwable) { bytes.fill(0); throw t }`. Pinned by a new JVM test (`databaseKeyRetainAcrossZerosBufferIfBlockThrows`) that throws inside the block and asserts the master buffer is zeroed before the exception escapes. `Throwable` (not `Exception`) on purpose so an `Error` from JNI - native-library mismatch, OOM during nativeKey - still zeros; baseline updated for `TooGenericExceptionCaught` to match.

Caller-side semantics unchanged: `SqlCipherPassRepository.create` already catches the propagated throw at line 252 and returns `Failure(DatabaseCorrupt)`.

**Not addressed in this PR (follow-up: wpass-t5g):** the reviewer's #2 (two methods on one buffer is a footgun) and #3 (`var`/`!!` dance). Their proposed `RetainedKeyBuffer` shape is the right direction for type-level enforcement of the lifetime contract; their exact snippet leaks the master buffer (comment "Master buffer untouched"), so the follow-up amends it to zero the master after copy. Holding that refactor for a separate PR rather than expanding this one's scope.

PR #52, opened 2026-05-07
passes-storage: route cipher_compatibility through SQLiteDatabaseHook (wpass-aio)

bittelc:
## Summary

Third attempt at fixing wpass-aio. The first two (#49 logcat logging, #51 retainAcross / keyHandle close) didn't address the real failure mode. This one does.

**Bug:** Reopen of an existing v4-on-disk SQLCipher DB fails with `SQLiteOutOfMemoryException` (SQLITE_NOMEM) from page-1 decrypt at `readSchemaVersionIfPresent`. Surfaces in the wallet UI as "Passes unavailable - Secure storage is unavailable."

**Root cause (per bd notes on wpass-aio):** `PRAGMA cipher_compatibility = 4` was applied *after* `SQLiteDatabase.openOrCreateDatabase`. On reopen, the binding has already chosen a default KDF regime by the time the post-open PRAGMA could fire; the lazy page-1 decode runs with the wrong KDF and the native `sqlite3_malloc` fails. Process #1 (create+write) masked the bug because there were no on-disk pages to decrypt yet - the post-open PRAGMA still landed before the schema DDL flushed page 1 with v4 KDF/MAC.

**Fix:**
- Move `cipher_compatibility = 4` into a `SQLiteDatabaseHook` (`preKey` + `postKey`) so the v4 KDF regime is locked before the first decode. Both hooks are belt-and-suspenders against binding builds that ignore one or the other for this pragma.
- Issue the pragma via `SQLiteConnection.execute(...)`, the only API available inside `preKey` (no `SQLiteDatabase` exists yet).
- Keep `PRAGMA foreign_keys = ON` post-open: it is connection-scoped and does not affect KDF or page format.
- File-level KDoc gains a maintenance note that any future KDF/page-format pragma MUST go through the hook.

**What is preserved from #51:** `retainAcross` + the keyHandle close path. Its pool-lifetime hypothesis was wrong about SQLITE_NOMEM (same-process create+write succeeded fully without it; new-process reopen cannot benefit from a "retained" buffer that was never built in process #1 JVM), but it does fix the closeable-leak warning from the original bug report. The cleanup of the `var openedDb` / `!!` awkwardness is tracked separately as wpass-t5g.

## Test plan

- [x] gradlew :passes-storage:detekt - passes
- [x] gradlew :passes-storage:testDebugUnitTest - passes (includes the existing PublicApiSurfaceTest from #51)
- [x] gradlew :passes-storage:compileDebugAndroidTestKotlin - new test compiles
- [ ] **Device test (required to confirm fix)**: on FP4 / Android 16 with waltDebug:
  1. pm clear is.walt.app
  2. Install the build
  3. Open Passes screen - should NOT show "Secure storage is unavailable"
  4. Import golden.pkpass, observe passes_pass_rendered
  5. am force-stop is.walt.app
  6. Relaunch and tap Passes - observe NO passes_storage_failure; the imported pass renders
- [ ] **Instrumentation test on emulator/device**: gradlew :passes-storage:connectedDebugAndroidTest should run the new CipherCompatReopenTest along with the existing instrumentation suite. Same-process fresh-instance is the closest the JUnit harness gets to "fresh JVM"; the SQLCipher binding per-connection KDF state lives on the SQLiteDatabase handle, so closing process #1 repo and constructing a new one is sufficient to reproduce the open-time KDF mismatch.

### DIFF
--- a/passes-storage/src/androidTest/kotlin/is/walt/passes/storage/CipherCompatReopenTest.kt
+++ b/passes-storage/src/androidTest/kotlin/is/walt/passes/storage/CipherCompatReopenTest.kt
@@ -31,17 +31,21 @@ import java.security.KeyStore
  * `SQLiteDatabase.openOrCreateDatabase`, which left the binding free to pick a non-v4
  * default KDF on reopen. The lazy page-1 decode then ran with the wrong KDF and the
  * native `sqlite3_malloc` failed, surfacing as `SQLiteOutOfMemoryException` (SQLITE_NOMEM)
- * from `readSchemaVersionIfPresent`'s schema-probe SELECT. Process #1 (create+write)
- * masked the bug because there was nothing on disk to decrypt yet.
- *
- * The fix is to issue the pragma from a [net.zetetic.database.sqlcipher.SQLiteDatabaseHook]
- * preKey/postKey so the KDF regime is locked before the binding decrypts page 1; this
- * test fails on the pre-fix code path and passes once the hook is installed.
+ * from `readSchemaVersionIfPresent`'s schema-probe SELECT.
  *
  * Same-process-fresh-instance is the closest the JUnit harness gets to "fresh JVM";
  * the SQLCipher binding's per-connection KDF state lives on the [SQLiteDatabase] handle,
  * so closing process #1's repo and constructing a new one is sufficient to reproduce
- * the open-time KDF mismatch.
+ * the open-time KDF mismatch on devices whose binding default differs from v4.
+ *
+ * Scope: green-path only. Asymmetric-by-design — there is no negative test that opens
+ * without the hook and asserts a failure, because the failure mode is
+ * binding-default-dependent: the bug deterministically reproduces on FP4 / Android 16
+ * with sqlcipher-android 4.6.1, but not necessarily on every CI emulator (a binding
+ * whose runtime default already aligns with v4 would pass a no-hook reopen). A negative
+ * test would be flaky-by-device, just inverted. The protection against accidental hook
+ * removal lives in code review and the maintenance KDoc on
+ * [is.walt.passes.storage.internal.SqlCipherDatabaseFactory], not here.
  */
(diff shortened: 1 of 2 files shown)


bittelc:
Pushed fb82f8a in response to the review. Per-point:

**1. preKey override:** Dropped. Kept a no-op preKey to satisfy the interface and added a comment pointing at the postKey placement rationale. Hook KDoc rewritten to state the Zetetic-documented placement instead of hedging.

**2. StorageResult contract for hook throws:** Wrapped the openOrCreateDatabase call in try/catch shaped like the schema-probe and DDL/migration blocks. Extracted to a new private helper `openHandleWithKey` so openOrCreate stays readable; the `openedDb!!` is now scoped to that helper, where it is unreachable on the throw path. Hook-thrown exceptions now map to `StorageError.Unknown(DatabaseCorrupt)` via a new `openCorrupt` helper.

**3. Test scope:** Took the second option from the review (accept green-path-only, document explicitly). The CipherCompatReopenTest KDoc now states that the negative case is intentionally not pinned because the failure mode is binding-default-dependent: a no-hook reopen would be flaky-by-device, just inverted. Pointed at the SqlCipherDatabaseFactory maintenance KDoc as the protection against accidental hook removal in lieu of a CI assertion.

**Smaller obs:**
- Shrunk the two chunky in-method comment blocks to one-liners that point at the file-level KDoc, per your last bullet.
- Internal-symbol coupling in the test (`WrappedKeyStorage.PREFS_NAME`, `AndroidKeystorePassKeyProvider.MASTER_KEY_ALIAS`): noted but left as-is since the same coupling already exists in `KeyUnavailableAcrossDataClearTest` and you flagged it as not-actionable-now.
- The hook is still an anonymous `object : ...` assigned to a val (your style-preference observation, no change).

Quality gates clean: `./gradlew :passes-storage:detekt :passes-storage:testDebugUnitTest :passes-storage:compileDebugAndroidTestKotlin`.

PR #56, opened 2026-05-12
passes-pdf: PdfImporter facade with API-34 gate, memfd, thumbnail (wpass-avs)

bittelc:
## Summary

- Adds the single PDF import entry point that ADR 0005 G.1 calls for, closing the last load-bearing gap under epic `wpass-i2r`. Without it, the consumer (walt-android) would have to compose the import sequence itself, which is the parallel-implementation pattern the repository's "DECISIVE CONSTRAINT" forbids.
- Implements the eight-step orchestration: API-34 SDK gate → bounded memfd-backed materialization → header sniff → memfd PFD allocation → bind isolated renderer service → probe → page-zero render → thumbnail PNG → consumer-supplied `persist` callback. Renderer service is unbound in `finally` regardless of outcome.
- Adds `DocumentRejectedKind.UnsupportedAndroidVersion` for the G.1 gate (with the wire-format mapping in `RejectedKindWire`), updates the surface-locked enum tests, and pins the new wire code at `5`.
- Storage handoff is a callback rather than a `PassRepository` dependency, so `passes-pdf` and `passes-storage` remain independent peers per the project's module rules.

## Files

- `passes-pdf-core`: `DocumentRejectedKind` gains `UnsupportedAndroidVersion`; `PublicApiSurfaceTest` updated to six arms.
- `passes-pdf` (new):
  - `PdfImporter` interface + `PdfImporter.create(context, config)` factory.
  - `PdfImportSource` sealed interface (`ContentUri`, `FileDescriptor`).
  - `DefaultPdfImporter` orchestrator with internal `Deps` seam for tests.
  - `internal/PdfPfdFactory` (`Os.memfd_create` → `ParcelFileDescriptor.dup`).
  - `internal/RendererSession` (`Context.bindService` → `PdfRendererClient`, with FailedSession fallback).
  - `internal/ThumbnailEncoder` (SharedMemory → Bitmap → PNG).
- `passes-pdf` tests:
  - `PdfImporterTest` (Robolectric, 14 cases) covers SDK gate, header sniff short-circuit, fail-fast size cap, every rejection arm, unbind on every outcome, persist-exactly-once, telemetry shape, ContentUri vs FileDescriptor sources, plus a reflection lock on the importer surface.
  - `PdfImporterInstrumentedTest` stubs the on-device coverage (benign-PDF end-to-end and the malformed-PDF rebind path), `@Ignore`d at check-in pending the same fixture set as `PdfRendererServiceInstrumentedTest`.

## Test plan

- [x] `./gradlew check` — all unit tests + lint + detekt + ktlint green across all seven modules.
- [x] 14 `PdfImporterTest` cases pass on Robolectric (SDK 34).
- [x] `RejectedKindWireSurfaceTest` and `passes-pdf-core::PublicApiSurfaceTest` pass with the new arm.
- [ ] Instrumented end-to-end (gated on the `wpass-5v9` fixture-set follow-up).

## Trust-claim notes

- Header sniff runs on the materialized bytes *before* `bindService`; the test `headerSniffShortCircuitsBeforeBindingService` asserts the recording session factory's `connectCalls == 0` for non-PDF input.
- Memfd path keeps PDF bytes off disk (ADR 0005 F.1). The `MemfdPfdFactory` uses `Os.memfd_create` + `Os.write` + `ParcelFileDescriptor.dup`; no `java.io.File`/`FileOutputStream`/`createTempFile` in the renderer-handoff code path.
- Telemetry stays enums-and-durations-only via the existing `DocumentTelemetryGuard`; no `String`, `ByteArray`, or `Throwable` parameters added.
- `unbindService` runs in a `finally` regardless of outcome — covered by `unbindRunsOnEveryRejectionAfterSuccessfulConnect` and the success/persist-throw cases.

## Unblocks

- `wpass-i2r` (PDF document support epic) — every kernel-side trust claim under it now lives in this repo.
- `wlt-4pg` (cross-repo wiring in walt-android).

## Cross-module impact

- `RejectedKindWire` adds code `5 = UNSUPPORTED_ANDROID_VERSION`; `RejectedKindWireSurfaceTest` updated; the renderer service binder never actually carries this code (the gate fires before the service is bound), but the wire-format table stays exhaustive.

### DIFF
--- a/passes-pdf-core/src/main/kotlin/is/walt/passes/pdf/PdfDocument.kt
+++ b/passes-pdf-core/src/main/kotlin/is/walt/passes/pdf/PdfDocument.kt
@@ -57,6 +57,14 @@ public enum class Provenance {
  *    `android.graphics.pdf.PdfRenderer` is not available. Fired by the importer entry
  *    point *before* any source bytes are read or the renderer service is bound; the
  *    isolated process never starts on a device that cannot satisfy the version floor.
+ *  - [EncoderFailed] → SharedMemory mapping, `Bitmap.copyPixelsFromBuffer`, or
+ *    `Bitmap.compress(PNG)` threw after the renderer service returned `Ok`. This is a
+ *    *post-renderer* failure inside the importer's PNG-encoding step. Distinct from
+ *    [RendererFailed] so telemetry can tell "PDFium choked on this file" apart from
+ *    "the device ran out of RAM during PNG encoding."
+ *  - [StorageHandoffFailed] → the consumer-supplied `persist` callback threw after a
+ *    successful render. Trust band is the storage layer (downstream of this module);
+ *    a spike here points the consumer at SQLCipher / DB infra rather than the renderer.
  *
  * Reviewers should treat any future addition of a string-bearing failure arm (e.g. an
  * "ErrorMessage" data class) as a security-policy change.
@@ -75,4 +83,6 @@ public enum class DocumentRejectedKind {
     TooManyPages,
     RendererFailed,
     UnsupportedAndroidVersion,
+    EncoderFailed,
+    StorageHandoffFailed,
 }
--- a/passes-pdf-core/src/main/kotlin/is/walt/passes/pdf/PdfDocument.kt
+++ b/passes-pdf-core/src/main/kotlin/is/walt/passes/pdf/PdfDocument.kt
@@ -52,6 +52,11 @@ public enum class Provenance {
  *  - [Encrypted] → D6 (encrypted PDFs are rejected at import).
  *  - [RendererFailed] → the isolated renderer service (D3) returned an error or timed out
  *    during page-count probing; we never report the underlying decoder error string.
+ *  - [UnsupportedAndroidVersion] → ADR 0005 G.1 runtime gate. The host device's
+ *    `Build.VERSION.SDK_INT` is below 34, so the Mainline-backed PDFium reachable through
+ *    `android.graphics.pdf.PdfRenderer` is not available. Fired by the importer entry
+ *    point *before* any source bytes are read or the renderer service is bound; the
+ *    isolated process never starts on a device that cannot satisfy the version floor.
  *
  * Reviewers should treat any future addition of a string-bearing failure arm (e.g. an
  * "ErrorMessage" data class) as a security-policy change.
@@ -69,4 +74,5 @@ public enum class DocumentRejectedKind {
     Encrypted,
     TooManyPages,
     RendererFailed,
+    UnsupportedAndroidVersion,
 }
--- a/passes-pdf-core/src/test/kotlin/is/walt/passes/pdf/PublicApiSurfaceTest.kt
+++ b/passes-pdf-core/src/test/kotlin/is/walt/passes/pdf/PublicApiSurfaceTest.kt
@@ -31,7 +31,7 @@ class PublicApiSurfaceTest {
     }
 
     @Test
-    fun documentRejectedKindHasExactlyTheFiveListedArms() {
+    fun documentRejectedKindHasExactlyTheSixListedArms() {
         val all =
             setOf(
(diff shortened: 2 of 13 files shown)


bittelc:
Review responses on 02d1f26:

**Accepted: #1 (scheme allowlist).** `openSource` now rejects any URI whose scheme is not `ContentResolver.SCHEME_CONTENT` before calling `openInputStream`. `file://` (and any other scheme) returns null at the open step, which already maps to `NotAPdf`. New test `nonContentSchemeUriRejectsAsNotAPdfWithoutBindingService` pins it — asserts the recording session factory's `connectCalls == 0` so a deliberate `file:///data/...` URI cannot reach `bindService`.

**Accepted: #2 (CancellationException).** Both `runCatching` wraps (encoder and persist) now rethrow `CancellationException` in their `getOrElse` blocks. New tests `persistCancellationPropagatesAndPreservesStructuredConcurrency` and `encoderCancellationPropagatesAndPreservesStructuredConcurrency` assert the exception escapes `import()` instead of folding to a rejection. The session is still unbound in `finally` even when cancellation propagates.

**Accepted (two-arm variant): #3 + #5.** Added `DocumentRejectedKind.EncoderFailed` (codes 6) and `DocumentRejectedKind.StorageHandoffFailed` (code 7). Routing now:

- `RendererFailed`: bind/probe/render window, plus pfd-alloc and source-fd `dup()` failures (these are "preparing the renderer handoff" — keeping them here avoids a third arm while keeping `RendererFailed` semantically about "the renderer pathway failed").
- `EncoderFailed`: SharedMemory map / Bitmap reconstruction / PNG compress threw post-render.
- `StorageHandoffFailed`: `persist` callback threw.

Picked the two-arm option over the one-arm `ImportInfrastructureFailed` because each new arm represents a distinct trust boundary (our code vs. the consumer's storage layer) — that's the granularity the on-call needs to triage. Surface tests updated (`PublicApiSurfaceTest`: 6→8 arms; `RejectedKindWireSurfaceTest`: codes 6, 7 pinned). Existing `persistThrowFoldsToRendererFailedAndUnbinds` renamed and updated; new `encoderThrowFoldsToEncoderFailedAndUnbinds` added.

On #5: `dup()` failure still routes to `NotAPdf` via `BoundedRead.SourceUnavailable`. I considered routing it to `RendererFailed` (kernel resource exhaustion is closer to "renderer pathway broke" than "this isn't a PDF") and that's defensible, but the public-facing user message under `NotAPdf` is appropriate ("we couldn't read this file") and the on-call signal is preserved via telemetry duration + the `dup()` log site. Open to flipping it if you'd prefer.

**Accepted: #4 (SharedMemory leak).** `PngThumbnailEncoder.encode` now uses the outer-try/finally layout you suggested:

```kotlin
try {
    val buffer = render.sharedMemory.mapReadOnly()
    try { ... } finally { SharedMemory.unmap(buffer) }
} finally {
    render.sharedMemory.close()
}
```

A throw inside `mapReadOnly()` now still runs `sharedMemory.close()`. KDoc updated to call out why the nesting matters.

**Skipped (per non-blocker call): #6.** Sealed-interface bookkeeping stayed as-is. The three two-arm interfaces (`Materialized`, `PfdAlloc`, `BoundedRead`) are each destructured immediately at the call site, but I agree the shape is more ceremonial than the problem demands. Happy to do the `Result<X>` collapse in a follow-up if you'd prefer — flagged it as worth tracking separately rather than mixing into the review-response commit.

**Accepted: #7 (FailedSession boilerplate).** Extracted `RejectingBinder(kind: DocumentRejectedKind)` to `internal/RejectingBinder.kt`. `FailedSession` now collapses to a one-liner. Helper is internal and is plausibly useful for future test fakes as you noted; not exposed in the public surface.

`./gradlew check` green across all seven modules. 18 unit tests on `PdfImporterTest` (14 prior + 4 new for scheme rejection, encoder throw, persist cancellation, encoder cancellation).

PR #59, opened 2026-05-14
passes-core: make minified consumer builds verify pkpass signatures (wpass-3kv, wpass-at6)

bittelc:
Two distinct R8/minification bugs that both surface every Apple-signed pkpass as `Failed(SignatureCryptoFailure)` in a release build. Found in sequence — the walt-android team tested the first fix on-device and the second cause was still there.

## Bug 1 — package-relative cert resource lookup (wpass-3kv)

`AppleTrustAnchors.loadResource()` loaded the bundled `.cer` trust anchors with a **package-relative** `getResourceAsStream("certs/...")`. R8 repackages the class in a minified consumer build, so the relative lookup resolved against the wrong package, returned `null`, and `error(...)` threw.

**Fix:** resolve the `.cer` files by **absolute** classpath path (`/is/walt/passes/core/internal/certs/`). R8 relocates classes, not java resources, so the absolute path is stable across minification.

> Note: this was a real latent bug but, per the walt-android team's on-device testing, **not** the one causing the observed failure — that is Bug 2. The absolute-path fix is correct and worth keeping regardless.

## Bug 2 — R8 strips the BouncyCastle JCA provider mappings (wpass-at6)

`SignatureVerifier` holds its own `BouncyCastleProvider` instance and passes it to the BC CMS / cert builders. `BouncyCastleProvider`'s constructor registers its algorithms **reflectively**: `Class.forName("<pkg>.<Alg>$Mappings")` per algorithm family, and each `$Mappings` class registers the actual SPI implementations by string class name. R8 sees no direct references and shrinks them away (consumer `mapping.txt`: 0 surviving classes in `jcajce.provider.digest`, 1 in `symmetric`, 3 in `asymmetric`). The provider constructs but registers almost nothing, so `Signature.getInstance("SHA256withRSA", BC_PROVIDER)` and `CertPathBuilder.getInstance("PKIX", BC_PROVIDER)` throw inside `verifySignature`; the outer `runCatching` buckets it as `SignatureCryptoFailure`.

**Fix:** `passes-core` ships BouncyCastle keep rules in `META-INF/proguard/passes-core.pro` so each consumer does not have to rediscover this — keep `org.bouncycastle.jcajce.provider.**` and `org.bouncycastle.jce.provider.**` whole, names included (the lookups are by literal string name). This is the minimal-practical set: `asn1`/`crypto`/`math`/`cms`/`cert`/`operator` are kept by R8 reachability since they are referenced directly or transitively, so the keep does not defeat shrinking the unused crypto.

## Shipping mechanism

`passes-core` is a plain Kotlin/JVM `java-library` with no Android `consumerProguardFiles` DSL. R8/AGP auto-applies rule files found at `META-INF/proguard/*.pro` inside dependency JARs on the consumer classpath — `passes-core.pro` rides that channel.

## Changes

- `AppleTrustAnchors.kt` — absolute classpath resource path.
- `META-INF/proguard/passes-core.pro` (new) — AppleTrustAnchors pin + BouncyCastle provider keep rules.
- `AppleTrustAnchorsTest.kt` — regression tests: certs resolve at the absolute path via a package-independent classloader; the shipped `.pro` contains the cert-loader and BouncyCastle keep lines.
- `SignatureVerifier.kt` / `SECURITY-CERTS.md` — doc/comment updates reflecting absolute-path loading and the R8/keep-rules dependency.

## Verification

- `./gradlew :passes-core:check` — green (detekt, ktlint, all tests incl. the new ones).
- `:passes-core:jar` inspection — `META-INF/proguard/passes-core.pro` (with all four keep/dontwarn lines) and all five `.cer` files land at the expected paths.
- **Pending — consumer-side:** a JVM unit test cannot reproduce R8. End-to-end check is an `isMinifyEnabled = true` walt-android release build: import a known-good Apple-signed pkpass and confirm `signature_kind=AppleVerified`. The walt-android team offered to test a candidate.

Addresses wpass-3kv and wpass-at6 — leave both open until consumer-side verification confirms.

### DIFF
--- a/passes-core/src/main/kotlin/is/walt/passes/core/internal/AppleTrustAnchors.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/internal/AppleTrustAnchors.kt
@@ -30,14 +30,17 @@ import java.security.cert.X509Certificate
  * with a network footgun.
  */
 internal object AppleTrustAnchors {
-    private val BUNDLED_TRUST_ANCHOR_FILENAMES =
+    // `internal`, not `private`, so AppleTrustAnchorsTest consumes these directly
+    // instead of mirroring them — the test then exercises the real bundled set and
+    // the real resource path, and adding a future anchor cannot leave a stale copy.
+    internal val BUNDLED_TRUST_ANCHOR_FILENAMES: List<String> =
         listOf(
             "apple-root-ca.cer",
             "apple-root-ca-g2.cer",
             "apple-root-ca-g3.cer",
         )
 
-    private val BUNDLED_INTERMEDIATE_FILENAMES =
+    internal val BUNDLED_INTERMEDIATE_FILENAMES: List<String> =
         listOf(
             "apple-wwdr-g3.cer",
             "apple-wwdr-g6.cer",
@@ -93,8 +96,11 @@ internal object AppleTrustAnchors {
      * `AppleTrustAnchors` to a synthetic package — a relative lookup then resolves
      * against the wrong package, returns `null`, and collapses every Apple-signed
      * pkpass onto `Failed(SignatureCryptoFailure)`. R8 relocates classes, not java
-     * resources, so the absolute path is stable across minification. If this
-     * directory ever moves, update this constant and `AppleTrustAnchorsTest`.
+     * resources, so the absolute path is stable across minification.
+     *
+     * `internal` (not `private`) so `AppleTrustAnchorsTest` resolves against this
+     * exact constant rather than a hand-kept copy; if this directory moves, the
+     * constant moves with it and the test follows.
      */
-    private const val RESOURCE_DIR = "/is/walt/passes/core/internal/certs"
+    internal const val RESOURCE_DIR: String = "/is/walt/passes/core/internal/certs"
 }
--- a/passes-core/src/main/kotlin/is/walt/passes/core/internal/AppleTrustAnchors.kt
+++ b/passes-core/src/main/kotlin/is/walt/passes/core/internal/AppleTrustAnchors.kt
@@ -6,10 +6,13 @@ import java.security.cert.X509Certificate
 
 /**
  * Loads the Apple trust anchors and known WWDR intermediates bundled under
- * `passes-core/src/main/resources/is/walt/passes/core/internal/certs/`. Loaded once on
- * first access and held in [BundledCerts]. The fingerprints and provenance of every
- * certificate are documented in `certs/SECURITY-CERTS.md`; an auditor or downstream
- * security reviewer can verify the bundled set without running the parser.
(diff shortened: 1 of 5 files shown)


bittelc:
## Verified end-to-end (walt-android team, minified `waltRelease`)

A fresh-install minified `waltRelease` imports a real Apple-signed pkpass as `signature_kind=AppleVerified`, where it previously failed `SignatureCryptoFailure`. Both fixes (wpass-3kv absolute resource path, wpass-at6 BC keep rules) are in the build that passed.

`mapping.txt` before vs after — the keep rules took effect, and the `jcajce.provider.*` count recovered as predicted:

| | before | after |
|---|---|---|
| `org.bouncycastle.jcajce.provider.*` classes | 6 | 1518 |
| `org.bouncycastle.jce.provider.*` classes | 3 | 74 |
| `*$Mappings` registrar classes | ~13 | 113 |

wpass-3kv and wpass-at6 both closed. PR ready for review.

bittelc:
Thanks for the review. Addressed in `2b89263`:

**#2 — De-duplicate the cert constants.** Done as recommended. `RESOURCE_DIR`, `BUNDLED_TRUST_ANCHOR_FILENAMES`, and `BUNDLED_INTERMEDIATE_FILENAMES` are now `internal` (with explicit types, per strict explicit-API mode) instead of `private`. `bundledCertsResolveAtAbsoluteClasspathPath` consumes them directly — `AppleTrustAnchors.RESOURCE_DIR.removePrefix("/")` for the classloader form — so it exercises the real path constant and the real bundled set. The mirrored `CERTS_RESOURCE_DIR` / `BUNDLED_CERT_FILENAMES` test constants are gone.

**#3 — Stale `SECURITY-CERTS.md` checklist.** #2 makes the "Updating" checklist gap moot as you predicted: the test now follows the production lists automatically. Also fixed a line in the loading note I'd added earlier in that file — it referenced "the path constants in `AppleTrustAnchorsTest`", which no longer exist; it now points at `RESOURCE_DIR` as the single source.

**#1 — Missing behavioral regression coverage on the BC keep rules.** Agreed this is the real exposure, and agreed it can only be exercised under R8 in the consumer. Filed as **wpass-56z** (P2) — a minified instrumented test in walt-android that imports a known-good Apple-signed pkpass against a minified build and asserts `signature_kind=AppleVerified`. Noted in the issue that the test must live in walt-android and can be relocated to its tracker.

**Minor #2 / #4 (string-match brittleness, `contextClassLoader` fidelity nit).** Acknowledged as-is — the `.pro` string-match test is a "did someone delete the rule" tripwire and nothing more; wpass-56z is the real contract test. The `contextClassLoader` choice is deliberate: it's the assertion that the path needs no package context, which is the regression this test exists to catch.

`./gradlew :passes-core:check` green after the changes.

PR #71, opened 2026-05-15
passes-pdf-ui: pinch-zoom + pan in DocumentView (wpass-1wq, wpass-0nn)

bittelc:
## Summary

- Adds a per-page zoom surface inside `DocumentView`'s `HorizontalPager` so users can enlarge a PDF-embedded barcode (PDF417/QR/EAN) enough for a venue or gate scanner to read it.
- Gesture priority: two-finger pinch always consumed locally; single-touch drag consumed only when `scale > 1` (`Modifier.transformable` with `canPan = { scale > MIN_SCALE }`), so swipe-to-page at fit is unchanged. Double-tap toggles fit ↔ `DOUBLE_TAP_SCALE`. Translation clamped so the page cannot pan off the slot.
- Zoom state lives in the existing `remember(document.id, pageIndex)` frame — paging away resets to fit.
- `DocumentTrustCaption` stays in `DocumentView`'s `Column` row above the pager and is structurally outside the zoom transform (ADR 0005 D5 non-suppressible contract preserved).
- ADR 0005 addendum (Z.1-Z.6) records the policy decision, including: 4 MP per-bitmap cap (D7) unchanged, binder surface (D4) unchanged, trust caption stays docked (option a, not panned with the page).

Renderer-side sharpness at zoom is the follow-up bead `wpass-f4b` (zoom-aware re-render of the visible viewport within the unchanged 4 MP cap). This PR is correct against the existing fit-resolution bitmap — pixelated when zoomed past ~1.3-1.5x, but the gesture surface and policy land here so the consumer integration in walt-android `wlt-o72.4` can move forward in parallel.

Parent epic: `wpass-6ag`.

## Test plan

- [x] `:passes-pdf-ui:compileDebugKotlin` clean
- [x] `:passes-pdf-ui:testDebugUnitTest` green
- [x] `:passes-pdf-ui:check` green (lint, detekt, ktlint)
- [x] `:passes-pdf-ui:compileDebugAndroidTestKotlin` clean
- [ ] Run instrumented tests on a device — needs `:passes-pdf-ui:connectedDebugAndroidTest`:
  - `pinchToZoomDoesNotAdvanceThePagerAndKeepsTheTrustCaptionVisible`
  - `doubleTapOnThePageDoesNotAdvanceThePager`
  - `singleTouchHorizontalDragAtFitScaleStillAdvancesThePager`
- [ ] Smoke-test on a Pixel: pinch a real ticket PDF; verify barcode area enlarges; verify trust caption stays on screen at MAX_SCALE; verify single-finger pan within zoomed page does not advance the pager; verify double-tap toggle.

Cross-repo: consumer integration tracked as walt-android `wlt-o72.4`. The walt-android `DocumentDetailScreen` embeds `DocumentView` in a vertical scroll container; the screen-level scroll will need to defer to the embedded zoom/pan when `scale > 1` (nested-scroll or zoom-state-gated suppression). Z.6 in the ADR addendum notes this as the consumer's responsibility.

### DIFF
--- a/docs/adr/0005-pdf-document-support.md
+++ b/docs/adr/0005-pdf-document-support.md
@@ -362,8 +362,12 @@ constants live in `passes-pdf-ui/.../DocumentView.kt`:
 
 - `MIN_SCALE = 1f` (zooming out below fit is meaningless on a
   single-page surface).
-- `MAX_SCALE = 5f` (brings a typical barcode to slot-filling size).
-- `DOUBLE_TAP_SCALE = 2.5f` (double-tap toggle target).
+- `MAX_SCALE = 3f` (interim ceiling against the current fit-resolution
+  bitmap — at 5x, each source pixel becomes a 5x5 bilinear block and
+  the smearing crosses into "bigger but less scannable" territory for
+  phone-camera barcode decoders. Moves to 5f once `wpass-f4b` lands the
+  viewport-aware renderer so the upsampling cost disappears).
+- `DOUBLE_TAP_SCALE = 2f` (double-tap toggle target).
 
 Gesture priority with the enclosing `HorizontalPager` is the
 load-bearing UX interlock:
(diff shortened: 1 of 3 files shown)


bittelc:
Pushed fixes for #3 and #4 in c124df6:

- **#3 (test gap)**: Added `singleTouchHorizontalDragWhileZoomedDoesNotAdvanceThePager`. Pinches outward, then performs a single-touch left swipe in the same pager that the fit-scale test confirms WOULD advance, and asserts page 1 is still displayed. Pins the positive half of `canPan = { scale > MIN_SCALE }` — without it, replacing `canPan` with `{ true }` or removing it would silently regress.
- **#4 (interim MAX_SCALE)**: `MAX_SCALE` → 3f, `DOUBLE_TAP_SCALE` → 2f. ADR addendum updated with the rationale and a note that this moves back to 5f once `wpass-f4b` lands the viewport-aware renderer. Comment in `DocumentView.kt` records the same.

Skipping #1 and #2:

- **#1 (combine `pointerInput(Unit) { detectTapGestures }` + `transformable`)**: They coexist correctly today. `detectTapGestures` doesn't consume events on movement or second-pointer-down — it just bails internally — so `transformable` still sees pinches and drags. The suggested refactor is a cleanliness pass, not a correctness fix; happy to take it if we revisit this code, but landing it now would be churn without a behavioural improvement.
- **#2 (`rememberUpdatedState` around `scale`)**: `var scale by remember { mutableFloatStateOf(1f) }` makes `scale` a delegated property — the lambda captures the *state delegate* by reference, and every `scale > MIN_SCALE` read inside `onDoubleTap` calls the getter and reads `delegate.value` fresh. `rememberUpdatedState` is the workaround for capturing a plain parameter inside a long-lived effect; with Compose State the state object itself is the "always fresh" mechanism, so wrapping it would be a no-op double-indirection.

PR #79, opened 2026-05-16
passes-storage: hand SQLCipher a RetainedKeyBuffer copy (wpass-t5g)

bittelc:
## Summary

Refactor recommended by PR #51 review for **wpass-t5g**. Splits the two consumers of `DatabaseKey`'s backing buffer onto separate arrays so the type system forbids the footgun that produced wpass-aio.

- `DatabaseKey.copyForRetainedConsumer(): RetainedKeyBuffer` copies the bytes, zeros the master, returns a buffer the long-lived SQLCipher pool owns.
- `DatabaseKey.withBytes { ... }` is unchanged.
- A second call through either method now sees an all-zero buffer, so a future caller cannot accidentally re-hand the live master.
- `SqlCipherDatabaseFactory.openHandleWithKey` drops the `var openedDb` / `!!` / lambda dance: the factory calls SQLCipher directly and `close()`s the buffer on the throw path.

`OpenedDatabase.keyHandle` stays typed as `AutoCloseable` so `SqlCipherPassStore`'s contract is the minimal "close in the right order" surface; nothing downstream needs the bytes.

## Test plan

- [x] `./gradlew :passes-storage:test` (JVM)
- [x] `./gradlew :passes-storage:detekt :passes-storage:lintDebug`
- [x] `PublicApiSurfaceTest` now covers: master is zeroed on copy; `close()` zeros the copy; two consecutive copies returns zero on the second; mixing `withBytes` after `copyForRetainedConsumer` sees zeros.
- [ ] No instrumented test changes needed; `:passes-storage:connectedCheck` round-trip still gated by the existing `CipherCompatReopenTest` / `KeystoreBackingMatrixTest` suite.

### DIFF
--- a/passes-storage/src/main/kotlin/is/walt/passes/storage/PassKeyProvider.kt
+++ b/passes-storage/src/main/kotlin/is/walt/passes/storage/PassKeyProvider.kt
@@ -35,21 +35,28 @@ public interface PassKeyProvider {
  * form, and exposes byte access only via [withBytes] (synchronous borrow) or
  * [copyForRetainedConsumer] (lifetime-tied handoff).
  *
- * Both access methods consume the master buffer: [withBytes] zeros after the block
- * returns, [copyForRetainedConsumer] zeros after copying. A second call after either
- * returns an all-zero buffer, so a future caller cannot accidentally hand zeros to
- * SQLCipher by mixing the two (the wpass-aio symptom class).
+ * Each [DatabaseKey] is single-use: the first call to either access method consumes
+ * it, and any subsequent call throws [IllegalStateException]. Silently re-handing an
+ * already-zeroed master to SQLCipher is the exact wpass-aio symptom class
+ * (page-1 decrypt with all-zero key surfaces as `SQLiteOutOfMemoryException`), so the
+ * second hand-off must surface loudly at the call site, not as opaque corruption.
+ * Single-threaded by construction; callers must not consume from multiple threads.
  */
 public class DatabaseKey(private val bytes: ByteArray) {
+    private var consumed: Boolean = false
+
     init {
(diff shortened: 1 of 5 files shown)


bittelc:
All three review items addressed in 673f53c.

**1. Throw-path coverage for `keyBuffer.close()`** — added `passes-storage/src/test/kotlin/is/walt/passes/storage/internal/SqlCipherDatabaseFactoryTest.kt`. `openHandleWithKey` is now `internal` and takes an `opener: (File, ByteArray, SQLiteDatabaseHook) -> SQLiteDatabase` parameter with the production default unchanged. Two tests inject throwing openers — one a generic Exception (translates to `StorageError.Unknown(DatabaseCorrupt)`), one a `CancellationException` (rethrown) — and assert the captured key buffer is all-zero after the factory returns / propagates. No SQLCipher native lib loads.

**2. Fail fast on double consumption** — `DatabaseKey` now carries a `consumed` flag set at entry to either access method; second hand-off (any combination) throws `IllegalStateException("DatabaseKey already consumed")`. The two "second-call returns zeros" tests were rewritten to assert the throw, with symmetric coverage for `withBytes`-twice and `copyForRetainedConsumer`-after-`withBytes`. Also added a flip-before-block test that locks in "throwing consumer still claims the key" semantics.

**3. Tightened `OpenedDatabase.keyHandle` and `SqlCipherPassStore.keyHandle`** from `AutoCloseable` to `RetainedKeyBuffer`. `closeQuietly` / `databaseCorrupt` signatures kept on `AutoCloseable` per the reviewer's note (they only call `close()`).

Gates: `./gradlew :passes-storage:test :passes-storage:detekt :passes-storage:lintDebug` all green.

PR #99, opened 2026-05-19
passes-ui: fix 1D barcodes collapsing to a 2 px strip in ScannableCardView (wpass-0j1)

bittelc:
## Summary

- `ZxingBarcodeEncoder` emits 1D barcode matrices at the symbology's natural minimum — one module tall — so a Code128 / Code39 / EAN-13 / UPC-A bitmap is `(~200, 1)` px. `Image` defaulted to `ContentScale.Fit`, which preserved the ~200:1 intrinsic ratio and painted at full container width but only ~1-2 dp tall, leaving the rest of the slot as whitespace. QR escaped because its matrix is square.
- Switch `ScannableCardView` to per-format `ContentScale`: `Fit` for QR (non-square slots stay undistorted), `FillBounds` for the four 1D formats so the painter stretches vertically. `FilterQuality.None` is already set, so module edges remain sharp.
- Add a Robolectric Compose regression test asserting a 1D barcode's rendered bounds fill a `320×200` dp container (pre-fix this collapsed to ~2 dp).

The encoder is left untouched: its "natural minimum" output is the cheapest correct encoding, and the bug is purely the renderer's wrong `ContentScale` assumption.

## Test plan

- [x] `./gradlew :passes-ui:testDebugUnitTest` — full module unit tests pass, including new `ScannableCardViewLayoutTest`.
- [x] `./gradlew :passes-ui:detekt` — clean.
- [x] `./gradlew assembleDebug detekt` — all seven modules build clean.
- [ ] Visual smoke check on device: open a Code128 ScannableCard in walt-android's scan screen and confirm the barcode fills the slot.


bittelc:
Thanks for the careful read. Verified the main claim by reverting the `contentScale = ...` line and re-running the test — it still passed. You're right that `getUnclippedBoundsInRoot` reports the `Image` composable's bounds, which fill the slot under both `Fit` and `FillBounds`; only the painted pixels differ. Tried rewriting with `captureToImage().toPixelMap()` to sample a row near the bottom, but `forceRedraw` times out under Robolectric in this setup (no real frame loop). Per your "prefer deletion over a misleading guard" recommendation, dropped the test in `bd093ba`. Point 4 (parameterize over all four 1D formats) goes with it.

Also addressed in the same commit:
- Point 2: replaced `ZxingBarcodeEncoder.kt:82-98` with the symbol reference `ZxingBarcodeEncoder.writeMatrix`.
- Point 3: dropped the `(wpass-0j1)` pointer from the KDoc.

Open to revisiting the regression guard if an instrumented test ends up worthwhile here; for now the fix stands on its own.

PR #106, opened 2026-05-24
passes-ui: ScannableCardView opt-in payload caption (GH #102)

### LINKED ISSUE #102: [Request] Display number below code
It would be helpful to display the number below the code in case the barcode scanner at the point of sale is not working

bittelc:
Closes #102.

## Summary

When the POS scanner cannot read the code, the user needs a human-readable fallback they can read aloud or paste. Adds `showPayloadCaption: Boolean = false` to `ScannableCardView`; when on, the encoded payload renders below the barcode as a monospace, user-selectable caption (`SelectionContainer` so long-press-to-copy works). FSI/PDI isolated as defense-in-depth on the create-boundary Cf/Cc rejection.

`ScannableCardScreen` is the only opt-in site. `ScannableCardTile` (carousel) and `ScannableCardRowTile` (wallet list) keep the default off — both are identification-only previews where a legible caption would not fit.

## Accessibility

When the caption is on, the inner image's `contentDescription` is dropped so TalkBack does not double-announce the label. The screen title already speaks `card.label`; the new caption is the announce-worthy text. Replaces the previous `clearAndSetSemantics {}` wrapper on `ScannableCardScreen`, which would have zapped the caption from a11y if left in place.

## Trust posture

`docs/SCANNABLE_CARD_THREAT_MODEL.md` row 4 (bidi / control-character spoofing in payload preview) is extended to acknowledge the detail-surface caption as a second display surface for the same already-mitigated threat. Same C3 mitigation: payloads with Cf/Cc codepoints are rejected at the create boundary, and the UI layer wraps the verbatim payload in FSI/PDI as defense-in-depth.

`ScannableCardView` is not in `ComposableSurfaceLockTest` (the lock pins `ScannableCardScreen`, `ScannableCardTile`, `ScannableCardRowTile`, `ScannableCardTrustCaption`, none of which gain a parameter here). No lock bump needed.

## Test plan

- [x] `./gradlew :passes-ui:testDebugUnitTest` — 20/20 ScannableCardTrustSurfaceTest (16 prior + 4 new: full-screen verbatim render for QR + 1D, default-off on tile + row)
- [x] `./gradlew :passes-ui:detekt` clean
- [x] `./gradlew assembleDebug detekt` across all seven modules clean
- [ ] Visual smoke on device: open a Code128 ScannableCard from the home lane, confirm the payload renders below the barcode in the full-screen scan view and is long-press-copyable

## Beads

Closes wpass-3hr.

### DIFF
--- a/docs/SCANNABLE_CARD_THREAT_MODEL.md
+++ b/docs/SCANNABLE_CARD_THREAT_MODEL.md
@@ -195,13 +195,19 @@ defense-in-depth, mirroring `B3UrlConfirmSheet` and the
 ### 4. Bidi / control-character spoofing in payload preview — Spoofing
 
 **What.** During create-time URI preview (C4), the dialog renders the raw
-`payload` so the user can see what they typed. The same bidi/control-char
-attack applies to that display.
+`payload` so the user can see what they typed. The detail surface
+(`ScannableCardScreen`) also renders the payload as a human-readable caption
+below the barcode (GH #102 — fallback for when a point-of-sale scanner cannot
+read the code; opt-in via `ScannableCardView.showPayloadCaption`). The same
+bidi/control-char attack applies to both displays.
 
 **Mitigation.** C3: `passes-core` rejects payloads containing Cf/Cc codepoints
-*before* any preview is shown. The dialog never receives a payload that could
-spoof its surrounding chrome. The fallback "looks URI-shaped but unrecognized
-scheme" path goes through the same validator.
+*before* any preview is shown or any caption is rendered. Neither the dialog
+nor the detail-surface caption ever receives a payload that could spoof its
+surrounding chrome. The fallback "looks URI-shaped but unrecognized scheme"
+path goes through the same validator. The UI layer additionally wraps the
+caption in FSI (U+2068) / PDI (U+2069) isolates as defense-in-depth, mirroring
+the label-isolation discipline in threat #3.
 
 **Status.** Mitigated.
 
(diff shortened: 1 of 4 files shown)


bittelc:
Thanks for the review. Pushed 9153ae7 addressing 1, 3, and the first minor:

- **Rec 1 (duplicated rationale).** Collapsed the screen-side comment to one line; the a11y reasoning stays in `ScannableCardView` where the semantics decision lives.
- **Rec 3 (encoder-failure path untested).** Added `fullScreenRendersPayloadCaptionEvenWhenEncoderFails`. The validator prevents that state in production, so the test bypasses it via the internal primary constructor (kotlin-reflect) using the same 11-digit EAN-13 payload `BarcodeEncoderTest.ean13RejectsWrongLengthAtWriter` uses. 21/21 now passing.
- **Minor 1 (BarcodeImage takes whole card).** Changed to `ScannableFormat` directly; the helper only ever read `card.format`.

**Rec 2 (CAPTION_GAP magic number).** Leaving as-is. You called it out as non-blocking and consistent with the existing 24.dp / 16.dp literals in the file. Surfacing it through `PassesSemantics` would mean introducing a new spacing token block, which feels like a larger design commitment than this fallback caption justifies — happy to follow up if you'd rather make that move now.

**Minor 2 (KDoc length).** Left as prose paragraphs; matches the style of the existing KDoc on this composable.

bittelc:
Pushed 40ccc81:

- **Rec 1 (reflection break radius).** Switched the fixture to `callBy(map)` keyed by `KParameter` name, with an upfront `require` checking the arg-name set matches the constructor's parameter set. A future add / remove / reorder of `ScannableCard`'s primary constructor now surfaces as an explicit `unknown params: …` error at test time rather than silently feeding the wrong field.
- **Minor (6-line comment).** Took your suggested rewrite verbatim, two lines.

**Rec 2 (branch ahead by 1):** the previous push (9153ae7) had landed before this review came in; `git status` here shows `up to date with 'origin/scannable-payload-caption'`. Likely a fetch-lag artifact on your end. 40ccc81 is now pushed; PR is current.

21/21 tests still green.

Issue #75, opened 2026-05-16
Allow renaming of passes


LuminousYoYo:
At present unable to rename passes inside app e.g. joebloogs.pdf.pdf might be better renamed as Joe Loyalty Card.

bittelc:
Done! Released in 0.1.7!

PR #111, opened 2026-05-24
passes-{storage,ui}: add user_label override and trust-caption rule (wpass-7xa)

bittelc:
## Summary

- Adds `passes.user_label` nullable column (v4 -> v5 schema migration) and `PassRepository.updatePassUserLabel(id, label)` — the third and final slice of the rename-APIs epic (wpass-osf), companion to walt-android `wlt-4dn.3` / GH #75. Siblings already shipped: #109 (updateDocumentLabel), #110 (updateScannableCard).
- Lands **ADR 0007** as the canonical reference for the trust-caption rule: when a pkpass's user_label override is set, every UI surface presenting that pass MUST also present the signed `organizationName`. Caption anchors on `organizationName`, not `description` (issuer identity > artifact-kind summary). updated_at is NOT bumped on user_label change (mirrors `updateDocumentLabel` / `updateScannableCard`).
- Introduces `PassIdentityBlock` in `passes-ui` as the **kernel-owned Composable** that enforces the trust-caption rule structurally (D6). `PassFront` re-wires its eyebrow through it; walt-android's pkpass tile / lane / detail surfaces are required to route through it rather than reading `StoredPass.userLabel` directly.
- Trim-blank-to-null normalization, 100-char cap, and case-insensitive equality suppression (so a rename equal to the signed org collapses to a single line) all locked by tests.

## Trust-claim posture

The signed `pass.json` body in `pass_json` BLOB is unchanged — user_label is a side-channel column. `Pass` (passes-core) is unchanged; the override never reaches the parser. `StorageTelemetryGuard.onUserLabelUpdated(type, hadPriorLabel, clearing)` carries no PII; the label text never enters telemetry.

## Test plan

- [x] `./gradlew :passes-storage:testDebugUnitTest` — 110 tests pass (new: schema v4 -> v5 migration walk-up, user_label column shape, updatePassUserLabel set/clear/trim-blank/over-cap/cap-precedes-IV/unknown-id, no updated_at bump)
- [x] `./gradlew :passes-ui:testDebugUnitTest` — 172 tests pass (new: PassIdentityBlock null/override/equality-suppressed/localized cases, PassFront userLabel parameter routes through PassIdentityBlock, PublicApiSurfaceTest lock on PassIdentityBlock symbol, ComposableSurfaceLockTest PassFront param count 8 -> 9)
- [x] `./gradlew testDebugUnitTest` — all modules green
- [x] `./gradlew assembleDebug` — debug build green across all modules
- [x] `./gradlew detekt` — no issues
- [ ] walt-android consumer wiring (`wlt-4dn.3`): rename UI calls `updatePassUserLabel`, reads `StoredPass.userLabel`, routes display through `PassIdentityBlock` per ADR 0007 D5+D6

### DIFF
--- a/docs/adr/0007-user-label-trust-caption.md
+++ b/docs/adr/0007-user-label-trust-caption.md
@@ -1,0 +1,312 @@
+# ADR 0007: user-label side-channel override and trust-caption rule for pkpass
+
+- Status: Accepted
+- Date: 2026-05-24
+- Tracks: `wpass-7xa` (this ADR + implementation), `wpass-osf` (rename-epic
+  parent), walt-android `wlt-4dn.3` / GitHub `walt-app/walt-passes-android#75`
+  (consumer-side rename UX). Sibling slices already shipped:
+  `wpass-hhq` (`updateDocumentLabel`, PR #109), `wpass-lay`
+  (`updateScannableCard`, PR #110).
+- Decision context: extends ADR 0002 D3 (the `passes` schema) and ADR 0003
+  (passes-ui theming surface). Trust-claim posture inherits from the repo
+  charter: every trust-claim-bearing behavior lives in this repository.
+
+## Context
+
+Walt-android wants a "rename pass" affordance: a user who has imported a
+Delta boarding pass should be able to label it "Mom's flight home" in the
+wallet list. PR #109 and #110 shipped the analogous capability for the two
+sibling artifact types (documents, scannable cards). pkpass is the
+highest-risk of the three because its title is part of the signed payload
+and therefore part of the trust bond.
+
+Today, a pkpass's visible identity on the front of the card and in lane
+tiles is the signed `pass.json` `organizationName` (after pass.strings
+substitution; see `PassFront.kt:102`). That string is what binds the user's
+mental model of *who issued this pass* to the cryptographically verified
+issuer identity. A naive "rename" feature that lets the user overwrite the
+displayed title would decouple displayed identity from signed identity in
+exactly the same shape as the #69 verifier-name swap incident: the screen
+says one thing, the cryptographically attested issuer is another, and there
+is no on-screen affordance to tell users which is which.
+
+The siblings did not have this problem. Documents (`updateDocumentLabel`)
+have no signed identity at all — the consumer supplies the label at import.
+Scannable cards (`updateScannableCard`) have no signature either; the
+payload is user-generated. pkpass is alone in carrying a PKCS#7-attested
+issuer identity that the rename feature must not silently hide.
+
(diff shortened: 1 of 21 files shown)


bittelc:
Thanks for the review. Responses point-by-point; landed in `a13a132`.

**1. BiDi isolation — applied.** Both `userLabel` and `displayOrganizationName` now wrap with `passes-ui-core::isolated` (FSI/PDI). Mirrors `ScannableCardTile` / `DocumentTile`. Added a dedicated `passIdentityBlockFencesBothLinesWithBidiIsolation` test that asserts the explicit `⁨…⁩` form, mirroring `securitySheetUrlIsBidiIsolated`. Six existing exact-text assertions switched to `substring = true` to accommodate the fence.

**2. `maxLines`/`overflow` — applied.** Primary line: `maxLines = 2, overflow = Ellipsis` (sibling `ScannableCardTile` / `DocumentTile` shape for user-controlled labels). Eyebrow: `maxLines = 1, softWrap = false, overflow = Visible` — mirrors `ScannableCardTrustCaption`'s "trust caption never truncates" posture so the signed issuer identity stays visible under bounded tile chrome.

**3. `@Suppress("LongParameterList")` — rejecting.** Empirically necessary: this repo's `detekt.yml:48` sets `functionThreshold: 6` (not the default 7), and detekt fires on six params (verified by removing the suppress and re-running; failure: `LongParameterList - 6/6 - [PassIdentityBlock]`). Kept the suppress and added an inline `// detekt functionThreshold=6; trips on six declared params.` so a future reader sees the rationale without re-running detekt.

**4. `trimmedOverride!!` — applied.** Refactored to a single `takeIf` chain that builds `override` (null = no override case), then `if (override == null) return Text(...)` followed by the two-line render with `override` smart-cast as non-null in the branch.

**5. KDoc trim — applied.** Dropped the per-param `primaryColor` / `eyebrowColor` paragraphs; defaults are visible at the signature. Kept the three-case render contract and the trust-claim sentences (the contract surface).

**Minor:**
- `0.7f` literal: declined. Single local derivation; a named constant adds ceremony without changing readability for one site.
- `@Suppress("ReturnCount")` placement: declined. Cosmetic; current placement matches the project's annotation style at function level.
- Orphaned KDoc in `StorageResult.kt:123-127`: applied. Reordered so `PassUpdateRejectedKind` goes ABOVE its own KDoc and the "Stable telemetry-friendly enumeration" block reattaches to `UnknownStorageFailureKind`.

Local verification: `:passes-ui:testDebugUnitTest` (173 pass), `:passes-storage:testDebugUnitTest` (110 pass), `compileDebugAndroidTestKotlin`, `assembleDebug`, `detekt` — all green.

PR #118, opened 2026-05-29
PDF full-screen: fix zoom aspect-ratio stretch + consumer-styleable open affordance (wpass-qz2)

bittelc:
Closes the two child beads of epic **wpass-qz2** (PDF detail + full-screen view improvements, Walt consumer request `wlt-cjv` / `wlt-o72`). Both surface in `passes-pdf` / `passes-pdf-ui`. The two design decisions were confirmed by the walt-android consumer team (slot-over-tokens for the affordance; full renderer+UI fix for the aspect bug).

## wpass-fdh (P1 bug) — full-screen PDF stretches beyond native aspect ratio when zoomed

**Root cause.** The full-screen zoom *replacement* path distorted the page. `PdfRendererService.matrixFor`'s `SubRect` branch scaled X and Y independently (`setScale(widthPx/srcWidth, heightPx/srcHeight)`) to fill a fixed bitmap, and `ZoomableImage` drew that replacement with `ContentScale.Fit` + `fillMaxSize()`. Whenever a pinch/pan clamped the visible window against the page's letterbox edge, the region's aspect diverged from the slot's and the page stretched. The base full-page render was already uniform, which is why the distortion was intermittent.

**Fix — stretching impossible by construction.**
- Renderer rasterises a sub-rect at a single **uniform** scale into a bitmap sized to the region's own aspect (`subRectOutputDims`), still within the 4 MP cap (ADR 0005 D7). The honest output dims flow back through the existing wire fields — no new attack surface or capability in the isolated process.
- `ZoomableImage` places the (now aspect-faithful) replacement at the on-screen box the region occupies (`visiblePageBoxOnScreen`) instead of filling the slot; `laneBackground` shows where the viewport extends past the page edge. A double-tap now also clears the stale sub-rect so it can't be shown against a changed transform.

## wpass-emn (P2) — full-screen open affordance not consumer-styleable

The affordance was a full-width, square-cornered text banner with only label + colours themeable, forcing consumers wanting a different register (Walt's centred rounded pill with a leading glyph) to suppress the kernel banner and lose the kernel-provided page tap target.

- Adds a `fullScreenAffordance` composable slot to `DocumentView`, mirroring `FullScreenDocumentView`'s `closeButton` slot. Host owns shape, corner radius, icon, label, padding; the kernel keeps the page tap target and the non-suppressible `DocumentTrustCaption`.
- The affordance floats over the bottom-centre of the page region (matching the requested design) rather than docking below it.
- Default renders the existing neutral banner, so label-and-colour-only consumers are unchanged. No new `DocumentSemantics` tokens (keeps the brand-visuals-via-slot convention).

## Trust surface
- `DocumentTrustCaption` remains non-suppressible; the affordance slot cannot touch it.
- No share / export / print / open-with affordance added.
- `DocumentView` parameter count bumps 6 → 7; `DocumentSurfaceLockTest` updated with rationale.

## Testing
- New JVM unit tests pin the no-stretch invariant on both sides: `SubRectOutputDimsTest` (renderer, uniform scale) and `ZoomablePlacementTest` (UI box aspect == rendered region aspect), including the letterbox-edge case that triggered the bug.
- `:passes-pdf` and `:passes-pdf-ui` unit tests + `detekt` pass; instrumented sources compile.
- Instrumented (`androidTest`) suites — including a new custom-affordance test and the existing pinch/zoom coverage — run on-device in CI (no local device available).

Beads: wpass-fdh, wpass-emn (children of wpass-qz2)

### DIFF
--- a/passes-pdf-ui/src/androidTest/kotlin/is/walt/passes/pdf/ui/DocumentViewInstrumentedTest.kt
+++ b/passes-pdf-ui/src/androidTest/kotlin/is/walt/passes/pdf/ui/DocumentViewInstrumentedTest.kt
@@ -2,10 +2,12 @@ package `is`.walt.passes.pdf.ui
 
 import android.os.ParcelFileDescriptor
 import android.os.SharedMemory
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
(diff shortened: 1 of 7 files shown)


bittelc:
Thanks for the review — addressed in d6acd53.

**1. Default banner overlapping page content** — fixed, and you're right the "unchanged" claim was inaccurate. Made the float **opt-in**: `fullScreenAffordance` is now nullable.
- Host supplies one → it floats over the page bottom-centre (Walt's pill, which intentionally overlaps the page edge per the Figma).
- Omitted (default) → the neutral banner **docks below the page exactly as before**. Default consumers are genuinely unchanged and page content is never obscured.
- KDoc reworded to state the placement contract accurately (no more "unchanged" overclaim). New instrumented test `defaultBannerDocksBelowThePageAndDoesNotOverlapIt` pins `banner.top >= page.bottom`.

**2. Double-tap blur** — fixed now rather than deferred, since it's contained and restores a property my change otherwise dropped. The zoom-in branch fires `onZoomedRegionChanged(visiblePageSubRect(...))` explicitly so a sharp sub-rect is requested (the settle-driven `snapshotFlow` can't fire without a transform edge). `wpass-f4b` parity with the pinch path is restored.

**Minor — comment density** — trimmed the 5-6 line blocks in `DocumentView.kt` to the ≤2-line "why".

**Minor — a11y labels on tap targets** — pre-existing; filed as **wpass-puz** rather than ship a half-measure, since a proper `onClickLabel` wants a localized string routed through `DocumentSemantics` (a deliberate surface addition + surface-lock bump), not a hardcode. Added `Role.Button` + label there when that copy lands.

`:passes-pdf-ui` unit tests + detekt pass; instrumented sources compile (run on-device in CI).

PR #119, opened 2026-06-03
ScannableCardScreen consumer-fit: showLabel param + bound white backing (wpass-1wu)

bittelc:
## Summary

Two consumer-fit UI adjustments to `passes-ui`'s full-screen `ScannableCardScreen`, surfaced by Walt's generated-code detail screen (android epic `wlt-u2b`). Tracked here as epic **wpass-1wu** with children wpass-1wu.1 and wpass-1wu.2; both land in this single PR.

The **trust contract is invariant**: `ScannableCardTrustCaption` stays bottom-docked and non-suppressible (C2 in `docs/SCANNABLE_CARD_THREAT_MODEL.md`), and `ComposableSurfaceLockTest`'s parameter-surface assertion is updated in lockstep.

## wpass-1wu.1 — `showLabel` param (Walt wlt-tct)

`ScannableCardScreen` always rendered `card.label` above the barcode. Walt's generated-code detail screen renders its own **editable** self-title above the kernel surface, so the kernel label was a duplicate.

- Added `showLabel: Boolean = true`, gating **only** the top label `Text`.
- Default `true` keeps every existing caller unchanged; the consumer passes `showLabel = false`.
- `showLabel` **cannot** hide the barcode, the payload caption, or the trust caption.

## wpass-1wu.2 — bound the white backing to the code (Walt wlt-n5z)

The screen painted `.background(surface)` across the full `fillMaxSize` column, so in a tall slot a 240dp QR floated inside a large white field.

- The code now sits on a **content-sized rounded `Surface`** (white backing + quiet-zone padding); the rest of the surface is transparent so the host's brand gradient shows through.
- ZXing bakes the scan quiet zone into the matrix (`ZxingBarcodeEncoder`), so scannability is preserved. Barcode min sizes (`ScannableCardView.minRenderSizeDp`) and the QR/1D `ContentScale` split are unchanged.
- No new public parameter.

## Trust-contract / test changes

- `ComposableSurfaceLockTest`: `scannableCardScreen…` lock updated `2 → 3` user-visible params, with a comment pinning that a 4th param (anything that could hide the caption) breaches C2.
- `ScannableCardTrustSurfaceTest`: added `fullScreenHidesLabelWhenShowLabelFalse` and `fullScreenShowLabelFalseStillRendersTrustAndPayloadCaptions` — the latter pins that `showLabel = false` still renders the trust caption + payload caption.

## Two judgment calls for android-team review

1. **White-card corner shape**: used `RoundedCornerShape(16.dp)` + a 16dp internal quiet-zone margin so the card reads as intentional against a gradient. No theme token exists for this; happy to flatten or tokenize if preferred.
2. **Backing color**: kept `MaterialTheme.colorScheme.surface` (matches the original token) rather than forcing pure white. The barcode bitmap carries its own white background + ZXing quiet zone, so scannability does not depend on this; only the 16dp padding ring takes the `surface` color.

## Testing

- `./gradlew :passes-ui:testDebugUnitTest --tests ComposableSurfaceLockTest --tests ScannableCardTrustSurfaceTest` — green
- `./gradlew :passes-ui:detekt` — green

### DIFF
--- a/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
+++ b/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
@@ -1,12 +1,13 @@
 package `is`.walt.passes.ui
 
-import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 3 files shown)


bittelc (commented):
Reviewed from the consumer side (Walt android `ui-fixups`, epic wlt-u2b). Both kernel beads are fully addressed. (Leaving the formal approval to the maintainer — this is a consumer-side sign-off comment.)

**wpass-1wu.1 (showLabel / wlt-tct)** ✅ — `showLabel` is appended last (existing callers unchanged) and gates only the top label `Text`; barcode, payload caption, and trust caption are all outside the gate. Binds exactly to the consumer call `ScannableCardScreen(card, showLabel = false, modifier = …)`.

**wpass-1wu.2 (bound white backing / wlt-n5z)** ✅ — full-bleed `.background(surface)` is gone; the code now sits on a content-sized `Surface(RoundedCornerShape(16.dp))` with the rest transparent so the host gradient shows through. `minRenderSizeDp` and the QR/1D `ContentScale` split untouched.

**Trust contract** ✅ — `ComposableSurfaceLockTest` 2→3 with the C2 comment, and `ScannableCardTrustSurfaceTest` adds the meaningful guard that `showLabel = false` still renders the trust + payload captions.

Verified locally:
- Walt `:feature:passes:compileDebugKotlin` against this branch (`-PwaltPassesPath`) → BUILD SUCCESSFUL; the `showLabel = false` reference resolves and unblocks wlt-tct.
- `:passes-ui` `ComposableSurfaceLockTest` + `ScannableCardTrustSurfaceTest` → green.

Answers to the two judgment calls:
1. **Corner shape** — keep `RoundedCornerShape(16.dp)` + the 16dp margin as-is; matches Walt's rounded register, no need to tokenize for this batch.
2. **Backing color** — keep `colorScheme.surface` (don't force pure white). One cosmetic note: in dark theme the 16dp padding ring renders dark around the barcode's ZXing-baked white block, so the "white card" reads as a dark framed card in dark mode. Scannability is unaffected (quiet zone is in the matrix). Defensible as-is; forcing `Color.White` would fight the dark theme on the rest of that screen.

Minor, non-blocking: the n5z layout change (content-sized vs full-bleed backing) isn't regression-locked by a test — the two new tests cover showLabel + caption presence only. Expected for a pixel/layout property; just flagging it isn't pinned.

PR #125, opened 2026-06-04
Pure-JVM ZXing barcode symbol decode (wpass-zrt.4)

bittelc:
## Summary

Lands the symbol-decode step of the isolated barcode decode epic (wpass-zrt.4). Fills the `BarcodeSymbolDecoder` seam that wpass-zrt.3 left open, completing the codec→symbol path inside the `:barcodeDecoder` sandbox.

- New internal `ZxingBarcodeSymbolDecoder` backed by **`com.google.zxing:core`** (Apache-2.0, **zero native attack surface** — the deciding reason over a native decoder or ML Kit, which carries telemetry + a Play-Services dep and breaks the degoogled walt.is channel). zxing-cpp stays a sandboxed contingency only.
- **Symbology allowlist, not decode-everything**: `POSSIBLE_FORMATS` is pinned to exactly the `ScannableFormat` roster (Code128 / Ean13 / UpcA / Code39 / Qr). PDF417/Aztec et al. are not enabled; a format somehow outside the roster folds to `DecodeFailed(UnsupportedBarcodeFormat)`.
- Payload is returned **faithfully** and never interpreted here — classification/validation stay downstream in the consumer (`QrPayloadClassifier` / `ScannableCardInputValidator`).
- `BoundedBitmapDecoder` now pins `ALLOCATOR_SOFTWARE` so the analysed bitmap is pixel-readable (the default hardware bitmap refuses `getPixels`). The bitmap is analysed, never displayed, and never crosses the binder.
- Removed the `NotYetImplemented` placeholder now that the real decoder is wired into `BarcodeDecodeService`.

## Tests

`decodeLuminance` is `Bitmap`-free, so the decode contract is exercised purely on the JVM by round-tripping through ZXing's own `MultiFormatWriter`:
- all 5 roster formats encode→decode back to the exact `{payload, format}`;
- a genuine PDF417 symbol reads as `NoBarcodeFound` — proving the allowlist;
- a blank image returns `NoBarcodeFound`.

`./gradlew :passes-barcode:check` is green (unit tests, ktlint, detekt, Android lint).

## Out of scope (follow-ups)

- The instrumented on-device decode (`benignQrDecodesToPayloadAndFormat`) stays `@Ignore`d pending fixtures + the `:passes-barcode` device CI matrix — owned by **wpass-zrt.5** (security suite), which gates the epic ship.
- License/advisory note: `com.google.zxing:core` is Apache-2.0; no open advisory for the pure-JVM core at 3.5.4.

### DIFF
--- a/passes-barcode/build.gradle.kts
+++ b/passes-barcode/build.gradle.kts
@@ -27,6 +27,12 @@ dependencies {
     // module's public surface — so implementation, not api.
     implementation(project(":passes-isolation"))
 
+    // Pure-JVM symbol decode (wpass-zrt.4). com.google.zxing:core is Apache-2.0 and carries
+    // ZERO native attack surface — it runs only inside the isolated sandbox and never on the
+    // public surface, so implementation, not api. ML Kit was rejected (telemetry + Play
+    // Services dep); zxing-cpp stays a sandboxed contingency only.
+    implementation(libs.zxing.core)
+
     testImplementation(libs.junit)
     testImplementation(libs.truth)
     testImplementation(libs.kotlinx.coroutines.test)
--- a/passes-barcode/src/main/kotlin/is/walt/passes/barcode/android/BarcodeDecodeService.kt
+++ b/passes-barcode/src/main/kotlin/is/walt/passes/barcode/android/BarcodeDecodeService.kt
@@ -36,13 +36,14 @@ import kotlin.coroutines.cancellation.CancellationException
 public class BarcodeDecodeService : Service() {
     private val config: BarcodeDecodeConfig = BarcodeDecodeConfig()
     private val watchdog: DecodeWatchdog = DecodeWatchdog(config.decodeTimeoutMs)
+    private val symbolDecoder: BarcodeSymbolDecoder = ZxingBarcodeSymbolDecoder()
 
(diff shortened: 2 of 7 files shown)


bittelc:
Thanks for the review. Resolution:

**1. UnsupportedBarcodeFormat unreachable** — Agreed, it's a defensive guard, not a live path (the pinned `POSSIBLE_FORMATS` allowlist can only return roster formats). Clarified the KDoc and added an inline comment marking it as belt-and-suspenders. Comment-only (ae86d97).

**2. UPC-A reported as EAN-13** — Accurate ZXing trait; correctly out of scope for this PR. The synthetic writer→reader round-trip returns `UpcA`, but a real photo may differ. Recorded as a note on **wpass-zrt.5** to assert real-photo UPC-A format/payload on-device and decide on leading-zero normalization. No production change here.

**Minor — `ReaderException → NoBarcodeFound` test gap** — Deferred to the instrumented corpus (also noted on wpass-zrt.5); forging a checksum/format failure synthetically is awkward, as you noted.

**Minor — KDoc length** — Left as-is per your call (matches the module's inline-security-rationale house style).

`:passes-barcode:check` remains green.

PR #128, opened 2026-06-04
Land runtime sandbox-isolation probes + area-bomb fixture (wpass-6mg, wpass-zrt.5)

bittelc:
## What

Lands the three remaining items that gated **wpass-zrt.5**'s headline runtime sandbox-isolation assertion, tracked as **wpass-6mg**. These are the on-device companions to the static pins in `ManifestPermissionsTest` and `BarcodeDecodeBinderSurfaceTest`.

### 1. Runtime sandbox-isolation probe (`decodeProcessCannotReachAppDataOrKeystore`)
New test-only `isolatedProcess` probe service (`BarcodeSandboxProbeService`) declared in a new `src/androidTest/AndroidManifest.xml`, mirroring `BarcodeDecodeService`'s sandbox declaration (isolated, not exported, own process). The test binds it and, **from inside the isolated process**, attempts the privileged calls the decode UID must never make:
- reading the app's private sentinel file → blocked,
- touching the Keystore daemon (`containsAlias`) → blocked,
- opening a socket → blocked.

It also reports its UID so the test confirms the checks ran in an isolated UID, not the app's. This is the contractual go/no-go boundary inherited from walt-android wlt-58a.1.

### 2. Raw-transact parcel probe (`decodeServiceReturnsNoBitmapOrSourceBytesOverBinder`)
A raw `transact` against the real `BarcodeDecodeService`, inspecting the reply parcel to assert only the pure `{tag, payload, format}` result crosses back — no trailing data, total size well under any `Bitmap`/source-bytes payload.

### 3. Over-area decompression-bomb fixture (`overAreaImageRejectsWithImageTooLarge`)
A hand-crafted small-file/huge-canvas PNG (8000×8000 = 64 MP) that stays under the 12000px per-side cap but trips the 50 MP area cap, rejected at the header listener before allocation. Crafted as raw PNG bytes so the fixture is <1 KB rather than a ~200 MB allocation.

Also broadens the malformed corpus with 0-byte and truncated-PNG fail-closed cases (wpass-zrt.5 corpus bullets).

## Verification

- All **8** `BarcodeDecodeServiceInstrumentedTest` scenarios pass on a physical **FP4 (Android 15)**.
- JVM `:passes-barcode:check`, `ktlintCheck`, and `detekt` are green.
- CI's connected-tests matrix (API 28/31/34/36) runs the full suite — no `@Ignore`'d skeletons remain.

## Issues

Closes wpass-6mg. Unblocks and completes wpass-zrt.5 (runtime isolation assertion now real).

### DIFF
--- a/passes-barcode/src/androidTest/AndroidManifest.xml
+++ b/passes-barcode/src/androidTest/AndroidManifest.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  androidTest-only manifest for passes-barcode. Declares a test-only isolated probe service
+  used by BarcodeDecodeServiceInstrumentedTest.decodeProcessCannotReachAppDataOrKeystore to
+  run code INSIDE an isolated process and prove the sandbox cannot reach app data, the
+  Keystore, or the network (the runtime companion to ManifestPermissionsTest's static pin).
+
+  This service exists ONLY in the test APK; it ships no decode logic and never touches the
+  main manifest. It mirrors BarcodeDecodeService's sandbox declaration exactly — isolated,
+  not exported, in its own process — so the privileges it probes are the same ones the real
+  :barcodeDecoder process runs under.
+-->
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <service
+            android:name="is.walt.passes.barcode.android.BarcodeSandboxProbeService"
+            android:isolatedProcess="true"
+            android:exported="false"
+            android:process=":probe" />
+    </application>
+
+</manifest>
--- a/passes-barcode/src/androidTest/kotlin/is/walt/passes/barcode/android/BarcodeDecodeServiceInstrumentedTest.kt
+++ b/passes-barcode/src/androidTest/kotlin/is/walt/passes/barcode/android/BarcodeDecodeServiceInstrumentedTest.kt
@@ -1,8 +1,16 @@
 package `is`.walt.passes.barcode.android
 
+import android.app.Service
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
+import android.content.ServiceConnection
 import android.graphics.Bitmap
 import android.graphics.Color
+import android.os.IBinder
+import android.os.Parcel
 import android.os.ParcelFileDescriptor
+import android.os.Process
 import androidx.test.ext.junit.runners.AndroidJUnit4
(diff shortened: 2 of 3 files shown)


bittelc:
Thanks — all three recommendations were accurate and are addressed in f9af4c2.

**1. Network probe false-positive (the important one).** Correct: connecting to a non-routable address fails by timeout for *any* process, so the old check couldn't tell a sandbox denial from an unreachable host — a regression granting `INTERNET` would have stayed green. Switched to probing the **capability** directly: the `socket(AF_INET, SOCK_STREAM, 0)` syscall is gated on the `AID_INET` gid an isolated process lacks. Verified on the FP4 that the isolated process (uid `99402`) gets **`EACCES`** on that syscall, so `networkBlocked` is now tied to the actual permission, not a routing outcome. (First tried distinguishing `EPERM`/`EACCES` on a `connect()`, but Android wraps that denial without an `ErrnoException` cause; the direct syscall is the clean signal.)

**2. Ignored `transact` return.** Both raw-transact sites now use `check(binder.transact(...)) { "… not handled" }`, so an unhandled code fails clearly instead of reading a garbage parcel.

**3. Comment verbosity.** Trimmed the three flagged block comments to the load-bearing sentence per the ≤2-line guideline.

Minor points: agreed, no change needed — the sentinel is always created by the app first (so app-data failure is a genuine permission denial), and `"app-only-secret"` is a harmless literal.

All 8 `BarcodeDecodeServiceInstrumentedTest` scenarios remain green on the FP4; JVM `check`, ktlint, and detekt pass.

bittelc:
Heads-up: this latest review is against an earlier snapshot (commit `16609e9`) and was already addressed in `f9af4c2`, the current PR head. Quick mapping to current code:

1. **Network probe** — no longer a connect to a non-routable address. It now probes the *capability* directly: `socket(AF_INET, SOCK_STREAM, 0)`, gated on the `AID_INET` gid the sandbox lacks (`BarcodeSandboxProbeService.kt:60-66`). Verified the isolated process gets `EACCES` on that syscall, so it's tied to the permission, not a timeout — strictly stronger than the suggested fix.
2. **`transact` return** — both sites are wrapped in `check(binder.transact(...)) { … }` (`BarcodeDecodeServiceInstrumentedTest.kt:148, 183`).
3. **Comment verbosity** — the three flagged blocks were trimmed to the load-bearing sentence in `f9af4c2`.

A fresh review of `f9af4c2` should come back clean on these three. Happy to take another pass if anything remains.

Issue #74, opened 2026-05-16
Allow image upload


LuminousYoYo:
Currently attaching passes as images e.g. fjpeg from phone not allowed. This requires conversion to pdf to upload.

Allow image attachment.

bittelc:
This is addressed in 0.1.10! Thanks for filing the feature request!

PR #146, opened 2026-06-12
Create new modules for image-based passes

CPinelli:
Addresses https://github.com/walt-app/walt-passes-android/issues/74

Add modules to support image-based pass provisioning.
The modules are organized in the same fashion as the existing PDF modules.

## Testing

I created a local app that consumes the library. I compared the new image UI to the existing PDF UI, and added a similar trust string.

Tested on Android Studio's virtual device, using a Pixel 10 Pro (API 37)

Screenshot from 2026-06-12 19-02-26

Screenshot from 2026-06-12 19-02-31

## Notes

I created those new modules, similarly to PDF, however I think there's a little bit of abstraction that can be done, and some common helpers can be extracted out too.


bittelc:
Interesting! Thanks for the PR!

Could you help me understand the use case here? Now that I've enabled live camera scanning, and image upload scanning, I would think that the original need described #74 would be addressed.

CPinelli:
> Interesting! Thanks for the PR!
> 
> Could you help me understand the use case here? Now that I've enabled live camera scanning, and image upload scanning, I would think that the original need described #74 would be addressed.

The main difference is that some passes don't technically have a barcode, and I didn't think modifying the barcode module made sense, because it's a good way to validate a pass, since the barcode is decoded to validate it.

Some cards don't have a QR code (e.g. auto insurance cards, health insurance cards..., at least here, in the U.S.), yet they're fully available on Apple wallet / Google Wallet.
I'm not sure if this is out of scope for Walt, though. If you feel like it doesn't belong, that's totally valid!

Here's an example of an auto insurance card:
mock-wallet-pass

@bittelc Note that if you decide that this feature could be included, a smarter way to go with it might be to abstract out some code from the barcode modules into these new modules, detect whether there's a barcode or not, if not just accept the image, otherwise validate the barcode.
Let me know your thoughts!

bittelc:
Ok keep me going with the use case here. So you go to your progressive website on your phone, and you've got like, insurance or something with Progressive for your car. And you want to keep your insurance policy for your car in Walt, right? And Progressive enables this with Apple Pay and Google Wallet- Progressive has an app or button on their website once you've logged in that is "Add to Google Wallet". And then when you click that button their website/app, it adds it to the wallet. And in Apple/Google wallet, you can see your insurance policy number and your expiration date and stuff for the corresponding license plate.

Is that right?
Thanks for the help here, I've been building Walt Passes without having any experience beforehand of using passes within Google Wallet/Apple Pay, so it's been a learning process of how people use these apps.

CPinelli:
> Ok keep me going with the use case here. So you go to your progressive website on your phone, and you've got like, insurance or something with Progressive for your car. And you want to keep your insurance policy for your car in Walt, right? And Progressive enables this with Apple Pay and Google Wallet- Progressive has an app or button on their website once you've logged in that is "Add to Google Wallet". And then when you click that button their website/app, it adds it to the wallet. And in Apple/Google wallet, you can see your insurance policy number and your expiration date and stuff for the corresponding license plate.

That's exactly right! The best approach would be to have them onboard with Walt, but this would require partnerships with every insurance company. When they pass a PKPASS to Google / Apple wallets, they provide the same data they display on the app, it's just a shortcut to have it available quickly (i.e. without FaceID / auth).

One way to counter that behavior would be to take a screenshot of the insurance card, and add it directly into Walt. Note that it also works with a physical card too (depending on whether you go paperless or not), so you could also take a photo of your health insurance card, and add it directly into Walt too, which would not be possible, for now, because there's no code bar.

Basically the flow here is:
- Enroll with an insurance
- Receive a card (virtual or physical)
- Add that card to Google / Apple wallet
- Provide it to physicians, pharmacists, or law enforcement (in the case of a car insurance card)...

There might be some other cases I'm not thinking of, the idea here is more about convenience

bittelc:
Alright, I can kind of see this. Can you help me figure out what to do with this screen then? How might I organize this so that it's clear between all the different types of uploads/scans/manual-entry?
Screenshot_20260615-220947

CPinelli:
> Alright, I can kind of see this. Can you help me figure out what to do with this screen then? How might I organize this so that it's clear between all the different types of uploads/scans/manual-entry?


How would you feel about:

```
[ Add with a code ]
Use a photo, screenshot, or image of the pass

[ Add without a code ]
Use a photo, screenshot, or image of the pass
```

Then tapping either of those options opens an option picker:

```
[ Take a photo ]
Capture a physical pass

[ Choose from photos ]
Use an existing image or screenshot
```

I'm not a UI / UX person, but I think this is a common way to handle different sources. What do you think?

PR #152, opened 2026-06-16
Add an import / export option for Walt

### LINKED ISSUE #120: [Feature Request] Backup and Restore
Just a way to backup your existing passes and restore, incase of reset / new phone.

CPinelli:
### Fixes https://github.com/walt-app/walt-passes-android/issues/120

Wallet export / import
  
Introduce the passes-export-core and passes-export modules.

## Core modules

- `passes-export-core`: `ExportableArtifact` interface + `ArtifactKind` constants. Zero dependencies within the repo, both passes-core and `passes-pdf-core` depend on it (API) so domain objects implement the interface directly.
- `passes-export`:  `WalletExporter` / `WalletImporter` with two key-supply modes each: raw 256-bit key  and `PBKDF2-HMAC-SHA256` passphrase. `AES-256-GCM` envelope, outer JSON is readable without the key so the importer knows which credential to request before attempting to decrypt the file
- `ScannableCard` and `PdfDocument` implement `ExportableArtifact`, `StoredPass` does too (via `passes-storage`).
- `PassRepository.readAllForExport()` + `ExportSnapshot` give walt-android a single call to materialize all artifact types, including raw PDF bytes, for handoff to WalletExporter.
- Round-trip tested with a demo app, schema stamping, forward-compat unknown-key preservation, and authentication-failure detection

Each future types of passes (or possibly other artifacts) may extend `ExportableArtifact` to be included in the exported `.walt` file

### Testing (with demo app)

Failure (wrong passphrase):
Screenshot from 2026-06-15 21-14-18

Success (describing the file):
Screenshot from 2026-06-15 21-14-29


CPinelli:
No idea why `passes-export/src/main/kotlin/is/walt/passes/export/WalletExporter.kt` would be marked as a binary file by Github, it's viewable, but not directly from the UI, it would appear :thinking:

bittelc:
Ooo fun!
But I'm sorry, I didn't realize you were still working on new stuff @CPinelli - because one thing that I haven't been clear about is that the audits and internal pentests that I'm going through are very critical of anyone but employees (aka, only me) writing code that goes into codebase. So I'm not accepting PR merges into the repos even though they are open source.
The open source-ness is there to provide auditability and trust to customers. It is basically view only.

I'm going to update the README and Contribution files to make this more clear. Sorry I didn't do so earlier.

Thank you SO much for your attempted contributions though!

CPinelli:
>  The open source-ness is there to provide auditability and trust to customers. It is basically view only.

Oh, I got you. No worries :saluting_face:

PR #203, opened 2026-08-08
feat(passes-ui): add card-face tint to ScannableCardScreen (wpass-80y.1)

bittelc:
Kernel companion to walt-android **wlt-38v8** (Passes colour system redesign). Closes `wpass-80y.1`; unblocks `wlt-38v8.6`.

## What

`ScannableCardScreen` gains `faceTint: Color = Color.Unspecified` (5th parameter). `Unspecified` keeps today's `colorScheme.surface` face, so every existing call site is unchanged.

The tint reaches the **card face only**. The panel behind the code is now `SCAN_CODE_PANEL` — literally `Color.White` in both themes, mirroring `CompactCodeView`'s `COMPACT_CODE_BACKING` — because the code is real content and must stay theme-independent and scannable in dark mode.

## Notable decisions

- **`Color = Color.Unspecified`, not `Color? = null`.** Matches `PassIdentityBlock`'s existing sentinel precedent in this module and avoids boxing the value class.
- **Panel colour changed** from `colorScheme.surface` to literal white. This file's KDoc has claimed a "white backing" since wpass-1wu.2; the theme token contradicted it and left a dark quiet-zone ring around the code in dark mode.
- **Card body restructured** per the Figma frames (`4 · QR code` / `4b · Linear`): the label and the GH #102 payload readback now sit *on the face*, below the white panel, rather than above the card and inside the panel. Without this the tint renders as a thin ring and the label sits off-card.
- **Ink derived from tint luminance** (black/white) so an arbitrary consumer tint stays legible. That is a contrast guarantee, not a brand token — `passes-ui` carries no brand values. *Open question for wlt-38v8:* the frames specify ink `#171717` exactly; an exact match would need a second `faceInk` parameter. Flag before the 5-param lock settles.
- `ScannableCardView.showPayloadCaption` is no longer used by any kernel surface. Kept public (hosts composing their own detail surface) with KDoc updated.

## Out of scope (unchanged)

Persistence of which colour an item carries stays consumer-side (`WalletColorRepository`); the kernel never learns why a tint was chosen. `ScannableCard.color` stays removed (wpass-q5p). The trust-caption contract is untouched — neither `showLabel` nor `faceTint` can suppress the barcode, payload caption, or C2 caption.

The threat-model amendment arguing colour is trust-neutral is `wpass-80y.3`, and the contested `ScannableCardRowTile` accent is `wpass-80y.4`; neither is touched here.

## Tests

- Surface lock raised 4 → 5 with rationale.
- `SCAN_CODE_PANEL` pinned to `Color.White` (mirrors `CompactCodeViewTest`).
- A tinted render asserts barcode, label, payload, and the non-suppressible C2 caption all still display.
- The lock test's reflection helper now excludes synthetic `$default` bridges — value-class name mangling would otherwise surface `ScannableCardScreen-<hash>$default` as a second overload with an inflated parameter count.

Gates: `:passes-ui` `testDebugUnitTest` + `detekt` + `lintDebug` all green.

### DIFF
--- a/docs/SCANNABLE_CARD_THREAT_MODEL.md
+++ b/docs/SCANNABLE_CARD_THREAT_MODEL.md
@@ -352,10 +352,10 @@ defense-in-depth, mirroring `B3UrlConfirmSheet` and the
 
 **What.** During create-time URI preview (C4), the dialog renders the raw
 `payload` so the user can see what they typed. The detail surface
-(`ScannableCardScreen`) also renders the payload as a human-readable caption
-below the barcode (GH #102 — fallback for when a point-of-sale scanner cannot
-read the code; opt-in via `ScannableCardView.showPayloadCaption`). The same
-bidi/control-char attack applies to both displays.
+(`ScannableCardScreen`) also renders the payload as a human-readable readback
+on its card face, below the code panel (GH #102 — fallback for when a
+point-of-sale scanner cannot read the code). The same bidi/control-char attack
+applies to both displays.
 
 **Mitigation.** C3: `passes-core` rejects payloads containing Cf/Cc codepoints
 *before* any preview is shown or any caption is rendered. Neither the dialog
--- a/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
+++ b/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
@@ -128,8 +128,8 @@ private fun CodeCard(
     val face = if (tinted) faceTint else MaterialTheme.colorScheme.surface
     val ink = if (tinted) inkOn(faceTint) else MaterialTheme.colorScheme.onSurface
     // The payload is the POS fallback, not decorative meta: on a tint it takes full ink,
-    // because inkOn's worst case is already only 4.58:1 and any alpha would push the
-    // readback under WCAG AA. Untinted keeps the contrast-checked theme token.
+    // because INK_FLIP_LUMINANCE's worst case leaves no headroom for alpha. Untinted
+    // keeps the contrast-checked theme token.
     val payloadInk = if (tinted) ink else MaterialTheme.colorScheme.onSurfaceVariant
 
     Surface(
--- a/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
+++ b/passes-ui/src/main/kotlin/is/walt/passes/ui/ScannableCardScreen.kt
@@ -39,22 +39,15 @@ import `is`.walt.passes.ui.core.isolated
  * [CODE_QUIET_ZONE] adds visual breathing room inside the panel and the QR/1D
  * `ContentScale` split in [ScannableCardView] is unchanged.
  *
- * ## Face tint vs. code panel
- *
- * [faceTint] colors the card face only. The panel directly behind the code is
- * [SCAN_CODE_PANEL] — literally white in both themes, never a theme token and never the
- * tint — because the code is real content and must stay theme-independent and scannable
- * in dark mode. This mirrors `CompactCodeView`'s `COMPACT_CODE_BACKING` guarantee, so
- * both the list-face and detail-face renders of a code share one white-backing rule.
- * Routing the panel off that constant, or letting the tint reach it, is amending the
- * contract rather than refactoring it.
+ * [faceTint] colors the card face only; the panel behind the code stays
+ * [SCAN_CODE_PANEL], which carries that guarantee and the argument for it.
  *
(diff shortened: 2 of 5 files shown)


bittelc:
Review addressed in e3e631a. Verdict per item:

**Fixed as recommended**

- **1 — luminance threshold.** Confirmed: the black/white tie is at `sqrt(0.0525) - 0.05 ≈ 0.179`, and `#2E8B7F` (L ≈ 0.206) did take white at 4.1:1 over black at 5.1:1. Corrected to the tie point, which is also the worst case — nothing now scores below 4.58:1. The derivation is on the constant so the next person does not re-round it to 0.5.
- **2 — payload alpha.** Fixed, and the correction is stronger than proposed: since the corrected worst case is already 4.58:1, *any* alpha breaks AA somewhere in the tint range, so 0.87 would not be safe either. On a tint the readback takes full ink; it stays distinguishable by size and monospace face. Untinted keeps its theme token.
- **3 — duplicate TalkBack announcement.** Confirmed, and it was a regression against the exact rationale written three lines away. Extracted `ScannableCodeImage(card, imageDescription, modifier)` as suggested; `CodeCard` passes null. Public surface and overload lock untouched.
- **4 — test swatch hid 1 and 2.** Correct — `#BFEEEA` is L ≈ 0.78, the one branch that behaved. Added a contrast test over both palette rows, the extremes, and the mid-tones a 0.5 flip gets wrong. Verified it fails at the old threshold rather than trusting that it would.
- **6 — threat model.** Confirmed stale; Threat 4 now points at the readback `ScannableCardScreen` renders itself. Left `wpass-80y.3` separate: the epic sequences it deliberately, and its scope is the C1/C2 and ADR 0005 D1 amendment, not this line.
- **7 — repetition.** Consolidated onto `SCAN_CODE_PANEL` where the invariant is enforced.
- **Minor — inline `24.dp`** named `SCREEN_MARGIN`.

**Handled differently**

- **5 — alpha ignored.** Real, but documented on `@param faceTint` rather than enforced with `require`. Crashing a user's detail screen because a consumer passed a translucent tint trades a cosmetic mismatch for a hard failure in the field, and the kernel cannot composite correctly anyway — the host paint behind the transparent screen is not ours to read.
- **Minor — `!it.isSynthetic` vs `endsWith("$default")`.** Kept `isSynthetic`: it also covers value-class boxing bridges, which the name check would miss. Trimmed the comment instead, since the objection was that it needed explaining.

**Deferred, filed**

- **Minor — clipping at large font scale.** Real, and this PR made it worse by moving the label and payload inside the card. Filed as `wpass-9ps` rather than fixed inline: `verticalScroll` conflicts with the `weight(1f)` centring and must not detach the bottom-docked C2 caption, so it wants its own design pass plus a font-scale test.

**Not changed**

- **Minor — `faceTint = Color.White` hides the panel.** Degenerate input; no palette entry is white, and adding a border the frames do not show would cost every real tint to guard one unreal one.

`:passes-ui` `testDebugUnitTest` + `detekt` + `lintDebug` green on a forced full re-run.

bittelc:
Second review addressed in e52439f.

**Fixed**

- **1 — TalkBack dedup unpinned.** Correct, and the sharpest point in either round: `imageDescription = null` was the entire fix and the by-text assertions pass either way, because a `contentDescription` is not a text node. Added the assertion in both forms — raw and FSI/PDI-isolated — since the image would carry the raw label while the `Text` carries the isolated one. Verified it fails when `CodeCard` is swapped back to `ScannableCardView`, rather than trusting that it would.
- **2 — 4.58:1 stated twice.** Correct; it is derived from the threshold, so the copies drift. Cut from the `payloadInk` comment, which now points at `INK_FLIP_LUMINANCE` where the derivation lives.

**Not changed**

- **Minor — `ScannableCodeImage` outside the overload lock.** Declining: `ScannableCardView`, the public composable that actually renders the code, is not in that lock either. It guards chrome-bearing surfaces where an overload could add or drop a trust affordance, not the encode-and-paint primitives beneath them. Recommendation 1's test is also the stronger guard here — it pins the behavior directly rather than the shape that happens to produce it.
- **Minor — `faceTint = Color.White` hides the panel.** Holding, same reasoning as last round: no palette entry is white, and the frames show no border on the panel, so adding one would cost every real tint to guard an input the design never produces. Flagging it as a known consumer-side footgun in the bead rather than defending against it here.
- **Minor — `showPayloadCaption` has zero kernel callers.** Agreed and already tracked: `wpass-80y.1`'s notes carry it as an open question for wlt-38v8, to remove if no host takes it up. Not removing it in this PR, since that is a public-API deletion resting on a consumer answer we do not have yet.

`:passes-ui` `testDebugUnitTest` + `detekt` + `lintDebug` green on a forced full re-run.

PR #205, opened 2026-08-08
docs: record that colour carries no trust meaning (wpass-80y.3)

bittelc:
Kernel companion to walt-android `wlt-38v8`. Docs only; no code changes.

The 26.08.08 design revision decouples card colour from the issuer: colour means class, every artifact class gets a tint, and any item is user-reassignable from the detail surface. Three kernel docs encoded the opposite.

## `SCANNABLE_CARD_THREAT_MODEL.md`

New C1/C2 row, **colour carries no trust meaning**. The verified-band confusion class stays closed, from new premises:

1. No surface presents as verified, at list or detail scale.
2. Signature status is stated in words, on the detail surface only.
3. Colour is a per-class default the user can reassign, so it cannot encode provenance even accidentally - and this is said out loud, so a Bronze PDF is not read as a violation.

The row also bounds what `faceTint` grants (face only; cannot suppress barcode, payload readback, or trust caption; ink contrast derived and pinned; the kernel stores no colour) and what would need an amendment (colour derived from signature status, a colour field back on a kernel model, a tint reaching the code panel or page render).

Other edits in the same file:

- The wallet-row concession's condition 2 is marked **under review by `wpass-80y.4`** - its prohibition rested on the premise this PR removes, so `.4` can now decide it on the merits. The strip stays neutral until then.
- Threat 10 (colour picker as injection vector) updated: a picker exists again, consumer-side; the kernel takes a constructed `Color`, so there is no parse step to attack.
- Threat 1 status and the `wlt-mx2d` redesign row updated / marked partly superseded.
- Test matrix picks up the `wpass-80y.1` / `.2` tint pins.

## ADR 0005

New D1 addendum, two rows:

- **D1.C** withdraws the neutral-surface rule for documents and states the replacement argument (D5 never surfaced a signed indicator, so there is no ladder for a colour to imitate).
- **D1.L** is the amendment `wlt-o72.6` has been holding open since 2026-05-14: consumers MAY render documents inline in a unified wallet list, provided the sibling data-model contract holds, the row carries no verified affordance, and detail-surface provenance is unchanged.

## ADR 0003

The 2026-07-18 "issuer colors extend to the wallet-list pass card face" addendum is superseded. D4 itself is untouched - `PassFront` still applies `Pass.colors` on the pass surface - and the security-sheet family still takes no tint and must not.

Closes `wpass-80y.3`. Unblocks `wpass-80y.4`.

### DIFF
--- a/docs/SCANNABLE_CARD_THREAT_MODEL.md
+++ b/docs/SCANNABLE_CARD_THREAT_MODEL.md
@@ -92,18 +92,6 @@ homogeneous-list register and nothing else. A future consumer wanting a wallet
 list that also drops the detail-surface caption is amending this row, not
 filing a refactor.
 
-> **Dormant (wpass-80y, 2026-08-08).** The consumer no longer exercises this
-> concession: the 26.08.08 rules block quoted in the "colour carries no trust
-> meaning" row below says "No previews on list cards", and the redesign removes
-> the code tile, barcode band, image badge and photo thumbnail from every list
-> card face. The concession is not withdrawn - `CompactCodeView` stays, and a
-> host may still render an extracted code at list scale - but it is
-> unexercised, and condition 2's "neutral (never issuer-colored) card surface"
-> is replaced by the class tint that the "colour carries no trust meaning" row
-> grants. Conditions 1 and 3 govern
-> unchanged for any host that takes the concession up again, and the accepted
-> residual risk below applies only while it is exercised.
-
 **C1 / C2 — list-face code render concession (wpass-tjc.2; consumer epic
 wlt-mx2d).** The redesigned wallet list renders a scannable card's — and a
 composite artifact's extracted — ACTUAL code on its list card face, through the
@@ -138,6 +126,19 @@ recipient remains the authority on whether the code is credit-worthy (Threat
 9). A consumer wanting to render a list-face code without the condition-2
 eyebrow taxonomy, or wanting pkpass barcodes at list scale, is amending this
 row, not filing a PR.
+
+> **Dormant (wpass-80y, 2026-08-08).** The consumer no longer exercises this
+> list-face code render concession (the wallet-row concession above is
+> unaffected and stays live): the 26.08.08 rules block quoted in the "colour
+> carries no trust meaning" row below says "No previews on list cards", and the
+> redesign removes the code tile, barcode band, image badge and photo thumbnail
+> from every list card face. The concession is not withdrawn - `CompactCodeView`
+> stays, and a host may still render an extracted code at list scale - but it is
+> unexercised, and condition 2's "neutral (never issuer-colored) card surface"
+> is replaced by the class tint that the "colour carries no trust meaning" row
+> grants. Conditions 1 and 3 govern unchanged for any host that takes the
+> concession up again, and the accepted residual risk above applies only while
+> it is exercised.
 
 **Redesign context — where the C1 list distinction now lives (wpass-tjc.4;
 consumer epic wlt-mx2d).** The redesigned wallet list changes what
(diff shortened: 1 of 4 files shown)


bittelc:
Round two: all five review findings verified against the source and fixed in b23801c.

| # | Finding | Verified | Fix |
|---|---------|----------|-----|
| 1 | C2 pin table cites two non-existent tests | Yes - actual names are `…HasExactlyFiveUserVisibleParameters` (bumped by `wpass-80y.1`) and `…HasExactlyFourUserVisibleParameters` (since `wpass-2a2`); the latter also contradicted condition 1's "count: 4" in the same file | Both corrected, the `gv6` one with the bump noted inline |
| 2 | List-face code render concession left on withdrawn premises | Yes - the new row quotes "No previews on list cards" while the unmarked concession asserts codes render on list faces, and its condition 2 requires the neutral surface the new row withdraws | Marked **dormant**, not withdrawn: `CompactCodeView` stays and a host may take it up again under conditions 1 and 3 unchanged; condition 2's neutral-surface requirement is replaced by the class tint |
| 3 | ADR 0005 D5.T still cites the 10-param lock | Yes - `wpass-80y.2` renamed it to `…Eleven`, so one ADR had two tables naming the same lock differently | Ten → Eleven at both the table and the prose, with the later bump noted |
| 4 | Threat 10 heading still "n/a" | Yes | Now `Out of scope (consumer-side)`, matching rows 9 and 13 |
| 5 | "Bounded the same way" overstates parity | Yes - `documentFaceIsTinted` gates on `isSpecified && alpha > 0f`; `ScannableCardScreen` gates on `isSpecified` alone, so a transparent tint paints a transparent face and `inkOn` (luminance 0) returns white ink | Sentence scoped to the four rules that do hold for both; divergence recorded explicitly and filed as `wpass-80y.5`. Classified as a legibility hazard for a careless consumer, not a trust-signal hole - the trust caption survives either way, and both KDocs say "pass an opaque color" |

Minors also taken: ADR 0005's addendum normalized to `color` (the file had zero `colour` before this branch; the verbatim design quote and the threat-model row title keep their spelling so they stay greppable), ADR 0003 gains a tests-pinning table and `wpass-80y.2` on its Tracks line, and the stale `codePanelIsLiterallyWhiteNotAThemeTokenOrTheTint` reference in a `ComposableSurfaceLockTest` comment is fixed.

That last one is the only non-doc change in the branch, so CI now runs for real: `:passes-ui:check` passes locally.

bittelc:
Round three: three findings taken, one minor deliberately not taken. Fixed in 76201aa.

| # | Finding | Verified | Fix |
|---|---------|----------|-----|
| 1 | "Dormant" blockquote sits above the section it annotates | Yes, and it is the worst of the three: it followed the wallet-row concession's closing paragraph, and that concession is explicitly still live (condition 2 only "under review") | Moved below the list-face concession block. It also now names the concession outright and says "the wallet-row concession above is unaffected and stays live", so placement is no longer the only thing carrying the binding |
| 2 | D1.C claims a reciprocal cross-reference that runs one way | Yes - `D1.C` appeared nowhere in the threat model | Return leg added to "Bound of this row", next to the `wpass-80y.4` pointer |
| 3 | Security-sheet pin overclaims what a count lock proves | Yes on both halves: a count catches an added param but not a swapped one, and only the b3 lock enumerated its six | Row reworded to say exactly what the lock does and does not catch; the phone/email locks now enumerate their six the way b3 already did, so there is something for a reviewer to read against |

Minors: the mid-phrase wrap is rejoined, and the ADR 0005 param-count clause is split into the current lock plus a one-clause history.

**Not taken - row 10's heading separator.** Correct that it is now the only heading in the file using a hyphen where its siblings use an em dash. Leaving it: the repo owner has a standing no-em-dash rule for text written for them, and every other line this branch adds follows it. The earlier commit message said the heading "matches rows 9 and 13", which meant the *phrasing* ("Out of scope"), not the punctuation - fair catch on the wording being ambiguous. Happy to switch it to an em dash if the file convention should win here; that is the owner's call, not one to make silently in either direction.

